PR #2, opened 2018-08-23
Applying fixes using suggestions from FV

IssamT:
I will add some comments below to answer some questions in the comments

### DIFF
--- a/src/nodealgs/algevalnode.jl
+++ b/src/nodealgs/algevalnode.jl
@@ -1,4 +1,4 @@
-mutable struct SolsAndBounds ###FVC### why not simply struct
+mutable struct SolsAndBounds
     alg_inc_ip_primal_bound::Float
     alg_inc_lp_primal_bound::Float
     alg_inc_ip_dual_bound::Float
@@ -19,7 +19,6 @@ end
 function update_primal_ip_incumbents(incumbents::SolsAndBounds,
         var_val_map::Dict{Variable,Float}, newBound::Float)
     if newBound < incumbents.alg_inc_ip_primal_bound
-         ###FVC### should we use an epsilon to enforce a strict improvement
         incumbents.alg_inc_ip_primal_bound = newBound
         incumbents.alg_inc_ip_primal_sol_map = Dict{Variable, Float}()
         for var_val in var_val_map
@@ -45,7 +44,6 @@ function update_dual_lp_bound(incumbents::SolsAndBounds, newBound::Float)
         incumbents.alg_inc_lp_dual_bound = newBound
     end
 end
- ###FVC### do we assume a minimization problem, or do we want to make it generic for both min and max
 
 function update_dual_ip_bound(incumbents::SolsAndBounds, newBound::Float)
     new_ip_bound = ceil(newBound)
@@ -99,6 +97,13 @@ function update_alg_primal_lp_bound(alg::AlgToEvalNode)
     master = alg.extended_problem.master_problem
     primal_bnd = master.primal_sols[end].cost
     update_primal_lp_bound(alg.sols_and_bounds, primal_bnd)
+end
+
+function update_alg_primal_lp_incumbents(alg::AlgToEvalNode)
+    master = alg.extended_problem.master_problem
+    primal_sol = master.primal_sols[end].var_val_map
+    primal_bnd = master.primal_sols[end].cost
+    update_primal_lp_incumbents(alg.sols_and_bounds, primal_sol, primal_bnd)
 end
 
 function update_alg_primal_ip_incumbents(alg::AlgToEvalNode)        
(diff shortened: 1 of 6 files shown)


src/nodealgs/algevalnode.jl
```diff
@@ -1,4 +1,4 @@
-mutable struct SolsAndBounds ###FVC### why not simply struct
```
IssamT: The reason is that the attributes change values. if we don't put mutable all attributes are "final"

src/nodealgs/algevalnode.jl
```diff
@@ -19,7 +19,6 @@ end
 function update_primal_ip_incumbents(incumbents::SolsAndBounds,
         var_val_map::Dict{Variable,Float}, newBound::Float)
     if newBound < incumbents.alg_inc_ip_primal_bound
-         ###FVC### should we use an epsilon to enforce a strict improvement
```
IssamT: This indeed can improve the performance but we can't have a correct algorithm (even with a given tolerance on the objective value) using that. No matter how small the error we accept here it can accumulate to an error that we can not control.

src/nodealgs/algevalnode.jl
```diff
@@ -45,7 +44,6 @@ function update_dual_lp_bound(incumbents::SolsAndBounds, newBound::Float)
         incumbents.alg_inc_lp_dual_bound = newBound
     end
 end
- ###FVC### do we assume a minimization problem, or do we want to make it generic for both min and max
```
IssamT: We have though with Vitor to only support minimization. Do you think it is an issue?

src/nodealgs/alggeneratechildrennodes.jl
```diff
@@ -40,7 +40,7 @@ function retreive_candidate_vars(alg::AlgToGenerateChildrenNodes,
     frac_master_vars = Pair{MasterVar, Float}[]
     for var_val in var_val_map
         if typeof(var_val[1]) <: MasterVar
-            if !primal_value_is_integer(var_val[2],
```
IssamT: You're right. Done

src/problem.jl
```diff
@@ -68,9 +67,7 @@ end
 SimpleVarIndexManager() = SimpleVarIndexManager(Vector{Variable}(),
         Vector{Variable}(), Vector{Variable}(), Vector{Variable}())
 
-function add_in_var_manager(var_manager::SimpleVarIndexManager, var::Variable)
-         ###FVC### suggest to name it "add_var_in_manager"
```
IssamT: You're right. Done for both add_in_var_manager and add_in_constr_manager

src/problem.jl
```diff
@@ -1,24 +1,23 @@
...
+
+mutable struct ConstrMpFormStatus{C<:Constraint}
     constraint::C
     status_in_basic_sol::Int
 end
 
 mutable struct LpBasisRecord
     name::String
-    vars_in_basis::Vector{VarMpFormIndexStatus}
-     ###FVC### why not a list rather than a vector
```
IssamT: A vector is usually better stored in memory, unless there are operations of deletions and insertions very often. Do you think we will have those?

src/problem.jl
```diff
@@ -157,12 +154,10 @@ mutable struct CompactProblem{VM <: AbstractVarIndexManager,
 
     ### Current solutions
     obj_val::Float
-    obj_bound::Float
-    ###FVC### is it the obj_cutoff_val, the dual bound the primal incumbent ?
```
IssamT: We use the same names in MOI ObjectiveValue and ObjectiveBound. I have added a comment to explain it better.

src/problem.jl
```diff
@@ -157,12 +154,10 @@ mutable struct CompactProblem{VM <: AbstractVarIndexManager,
...
+    obj_bound::Float # Dual bound in LP, and "pruning" bound for MIP
     in_primal_lp_sol::Set{Variable}
     non_zero_red_cost_vars::Set{Variable}
-    ###FVC### unclear : do we mean strictly non-basic var ?
```
IssamT: It's not used for now. Maybe @rrsadykov can answer this.

src/problem.jl
```diff
@@ -329,8 +324,7 @@ end
 
 function sol_is_integer(sol::Dict{Variable, Float}, tolerance::Float)
     for var_val in sol
-        if (!primal_value_is_integer(var_val.second, tolerance)
-            ###FVC### this should be a geneic function testing if a float is integer; so the name primal_value is not appropriate
```
IssamT: You're right. Done

src/varconstr.jl
```diff
@@ -63,7 +63,6 @@ end
     # mutable struct = 'X' for constraints defining a subproblem convexity constraint
     #            in the master
     # ```
-     ###FVC### why vc-type and not just type
```
IssamT: type is a julia keyword, so we wanted to avoid the conflict.

src/varconstr.jl
```diff
@@ -137,7 +136,7 @@ end
     # problem setup-
     # ```
     treat_order_id::Int
-    ###FVC### better be called gen_sequence_number
+    # TODO better be called gen_sequence_number
```
IssamT: I agree those numbers aren't well names. There are 3 of them I think. I prefer however to keep the same name as in BaPCod for now to not get confused but we should change them in the future.

rrsadykov:
In non_zero_red_cost_vars we have variables with non-zero reduced costs, not all non-basic vars. This is used when resetting solution (to nullify reduced cost only for variables with non-zero reduced cost).

PR #8, opened 2018-09-06
Updates for MOI 0.6

IssamT:
Tests are passing on my mac. We need to wait for GLPK to be tagged, update the REQUIRE, and this should be good to merge.


vitornesello:
Great! Then we start working on this issue: https://github.com/ResourceMind/Coluna.jl/issues/7
?

PR #15, opened 2018-09-21
using JuMP utilities for MOI in demo app

IssamT:


demos/GeneralizedAssignment_ColGen/run_sgap.jl
```diff
@@ -1,4 +1,5 @@
 using JuMP
+using MathOptInterface: set
```
vitornesello: I don't think we should do this. We will lose consistency throughout the code. `MOI.` is ver short already

IssamT: The idea is to have JuMP applications written in a friendly language. that's why for example we don t write JuMP.@variable or JuMP.Model

vitornesello: Ok, thats only in the JuMP level

IssamT: Yes exactly and the MOI.set i use are the shortcuts defined in JuMP such as https://github.com/JuliaOpt/JuMP.jl/blob/2ae979eec4aeac1b6dc76d614b79c3c99c3dacc5/src/JuMP.jl#L468

PR #51, opened 2019-03-14
WIP : New Data Structure 

FD-V:
coluna


guimarqu:
We cannot have object `Constraint` that contains `Variable` and `Variable` that contains `Constraint` because julia does not support circular dependencies.
https://github.com/JuliaLang/julia/issues/269

vitornesello:
> We cannot have object `Constraint` that contains `Variable` and `Variable` that contains `Constraint` because julia does not support circular dependencies.
> [JuliaLang/julia#269](https://github.com/JuliaLang/julia/issues/269)

True, we can define AbstractVarConstr. Which will actually help us to define methods that apply to both variable and constraints.

src/constraints.jl
```diff
@@ -34,61 +32,11 @@ struct Constraint{DutyType <: AbstractConstrDuty}
     # Inactive = Can enter the formulation, but is not in it
     # Unsuitable = is not valid for the formulation at the current node.
     # ```
-    status::VCSTATUS
+    status
```
vitornesello: Why this is not typed anymore?

src/constraints.jl
```diff
@@ -1,13 +1,11 @@
...
+struct Constraint{T <: AbstractConstrDuty}
     uid::Int  # unique id
-    name::String
-    duty::DutyType
+    name::Symbol
+    duty::T
     formulation::Formulation
     vc_ref::Int
-    rhs::Float64
+    rhs::Float
```
vitornesello: Do you still want to keep this alias for `Float64`? I am not really comfortable with that. Every time I read the code I have to remember that we defined this alias, which is kind of weird because `Float` is an actual type in Julia

src/constraintduties.jl
```diff
@@ -0,0 +1,30 @@
+mutable struct Original <: AbstractConstrDuty
+    moi_def::MoiConstrDef # explicit Constr
+    rep_in_reform::Constraint # if any
+end
+
+mutable struct PureMasterConstr <: AbstractConstrDuty
+    moi_def::MoiVarDef # explicit var
```
vitornesello: `MoiVarDef` of `MoiConstrDef`?

src/constraints.jl
guimarqu: Because we don't need it yet. Moreover, the status could be only Active or Unsuitable. In that case, a Boolean should be sufficient.

src/constraints.jl
guimarqu: `Float` is not a julia type but we can remove the alias.

vitornesello:
Oh yes, you are right about not being a type. This just shows my point that it confuses me.

src/varconstr.jl
```diff
@@ -1,34 +1,31 @@
+abstract type AbstractVarConstr end
 abstract type AbstractVarDuty end
 abstract type AbstractConstrDuty end
 
-global const VarId = Int
+global const VarId = Int # maybe we should do a wrapper
 global const ConstrId = Int
```
guimarqu: Perhaps we should create an immutable `Index{T <: AbstractVarConstr}` that wraps the integer as they do in MathOptInterface https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/src/indextypes.jl

guimarqu:
It's possible to use a wrapped integer as id of the sparsevector but it will be a lot of work... 

Instead we should define the attribute `SparseVector{Float64, Int}` in the `VarMembership` type and write a method `getmembership(::VarMembership, ::VarIndex)`.


That's an example of what's going on when we want to instanciate a sparse vector using a set of homemade indices.
```julia
julia> struct X <: Integer
       x::Int
       end

julia> I = [X(1), X(4), X(10)]
3-element Array{X,1}:
  X(1)
  X(4)
 X(10)

julia> J=[9.0, 1.0, 3.0]
3-element Array{Float64,1}:
 9.0
 1.0
 3.0

julia> v = sparsevec(I, J)
ERROR: promotion of types Int64 and X failed to change any arguments
```

I started to do
```julia
Base.promote_rule(::Type{X}, ::Type{Int}) = Int
Base.convert(::Type{Int}, f::X) = f.x
Base.:<(a::X, b::X) = Base.:<(a.x, b.x)
```
and I still have an error
```
v = sparsevec(I, J)
ERROR: MethodError: no method matching sub_with_overflow(::X, ::X)
```

src/membership.jl
```diff
@@ -36,6 +36,33 @@ function get_constr_members_of_var(m::Memberships, uid::VarId)
...
+        end
+    end
+end
+
+function set_constr_members_of_var!(m::Memberships, uid::VarId, new_membership::ConstrMembership) 
+    m.var_to_constr_members[var_uid] = new_membership
+    constr_uids, vals = findnz(new_membership)
+    for j in 1:length(constr_uids)
+        if hasconstr(m, constr_uids[j]) 
+            m.constr_to_var_members[constr_uids[j]][var_uid] = vals[j]
```
guimarqu: Does not erase the membership of a variable that is now equal to zero. The name of the method also confuses me.

src/types.jl
```diff
@@ -0,0 +1,35 @@
...
+@enum Flag Static Dynamic Delayed Artifical Implicit
+@enum Status Active Unsuitable
+@enum ObjSense Min Max
+@enum SolutionMethod DirectMip DantzigWolfeDecomposition BendersDecomposition
+
+@enum VarDuty OriginalVar OriginalExpression PureMastVar MastCol MastArtVar MastRepPricingSpVar PricingSpSetupVar  PricingSpVar  PricingSpPureVar MastRepBendSpVar BendersSpVar BlockGenSpVar MastRepBlockSpVar
+
+@enum ConstrDuty OriginalConstr BranchingConstr MastPureConstr MasterConstr MastConvexityConstr PricingSpPureConstr  MasterBranch PricingSpRepMastBranchC
+
+const VarId = Int
```
vitornesello: Do you really think that these aliases are good? We will not know that `VarId` is just an `Int` by reading the code. I would rather wrap it in a `struct`. Same fo the other Ids.

src/constraint.jl
```diff
@@ -0,0 +1,101 @@
...
+    constr_uid::ConstrId
+    form_uid::FormId
+    name::String
+    rhs::Float64 
+    sense::ConstrSense # Greater Less Equal
+    kind::ConstrKind  # Core Facultative SubSytem PureMaster SubprobConvexity
+    flag::Flag    # Static, Dynamic/Delayed, Implicit
+    duty::ConstrDuty 
+    index::Union{MOI.ConstraintIndex, Nothing} 
+end
```
vitornesello: Shouldn't we type `Constraint` and `Variable` according to their Duties? For that we cannot use the enums defined in `types.jl`

``` julia
abstract type AbstractDuty end
struct SpVar <: AbstractDuty end
struct Variable{Duty<:AbstractDuty} <: AbstractVarConstr
       fields...
       duty::Duty
end
```

src/membership.jl
```diff
@@ -21,37 +21,53 @@ implicitmask(f::Filter) = f.used_mask .& f.implicit_mask
 explicitmask(f::Filter) = f.used_mask .& !f.implicit_mask
 #selectivemask(f::Filter, active::Bool, static::Bool, artificial::Bool) = f.used_mask active ? .& f.active_mask : nothing  static ? .& f.static_mask : nothing  artificial ? .& f.artificial_mask : nothing
 
+
+const VarMembership = Dict{VarId, Float64} #SparseVector{Float64, VarId}
+const ConstrMembership = Dict{ConstrId, Float64} #SparseVector{Float64, ConstrId}
```
guimarqu: We should avoid those shortcuts as much as possible because it may lead to duplicated code like the two functions that follow. I think the name of the variable should be clear enough to understand what it stores, and the type should describe how we store.

src/membership.jl
```diff
@@ -21,37 +21,53 @@ implicitmask(f::Filter) = f.used_mask .& f.implicit_mask
...
+function get_ids_vals(m::ConstrMembership)
+    #return findnz(m)
+    uids = Vector{ConstrId}()
+    vals = Vector{Float64}()
+    for (uid,val) in m
+        push!(uids, uid)
+        push!(vals, val)
+    end
+    return uids, vals
+end
```
guimarqu: duplicated code

src/membership.jl
```diff
@@ -21,37 +21,53 @@ implicitmask(f::Filter) = f.used_mask .& f.implicit_mask
...
+
 struct Memberships
     var_to_constr_members::Dict{VarId, ConstrMembership}
-    var_to_partialsol_members::Dict{VarId, VarMembership}
-    var_to_expression_members::Dict{VarId, VarMembership}
     constr_to_var_members::Dict{ConstrId, VarMembership}
+    var_to_partialsol_members::Union{Nothing,Dict{VarId, VarMembership}}
+    partialsol_to_var_members::Union{Nothing,Dict{VarId, VarMembership}}
+    var_to_expression_members::Union{Nothing,Dict{VarId, VarMembership}}
+    expression_to_var_members::Union{Nothing,Dict{VarId, VarMembership}}
```
guimarqu: An empty dict is sufficient

src/membership.jl
guimarqu: Why a Dict instead of a SparseVector ?

src/constraint.jl
guimarqu: We will do that if we need methods that dispatch variables according to their duty.

src/variable.jl
```diff
@@ -95,9 +97,24 @@ function reset!(v::Variable)
...
 
+get_cost(v::Variable) = get_cost(v.cur_data)
+get_lb(v::Variable) = get_lb(v.cur_data)
+get_ub(v::Variable) = get_ub(v.cur_data)
+get_kind(v::Variable) = get_kind(v.cur_data)
+get_sense(v::Variable) = get_sense(v.cur_data)
+get_inc_val(v::Variable) = get_inc_val(v.cur_data)
+is_active(v::Variable) = is_active(v.cur_data)
+is_explicit(v::Variable) = is_explicit(v.cur_data)
+set_cost!(v::Variable, cost::Float64) = set_cost!(v.initial_data, cost)
```
vitornesello: I don't think that these setters are a good idea because (same applies do constraint setters):
1. The caller does not know that they modifying `v.cur_data`
2. We would have to define the same setters for `cur_data`

PR #59, opened 2019-05-07
Update ci

guimarqu:


guimarqu:
Since `BlockDecomposition` is not a registered package, `Pkg`cannot resolve dependencies of `Coluna`. So we have to clone `BlockDecomposition` and then resolve. 

https://discourse.julialang.org/t/package-manager-resolve-complaining-of-unsatisfiable-requirements-due-to-no-known-versions/23778

PR #67, opened 2019-05-13
Strategies rework

vitornesello:
Implementation of the strategies philosophy last discussed.


src/strategies/strategy.jl
```diff
@@ -9,3 +7,22 @@ StrategyRecord() = StrategyRecord(StartNode, Dict{Symbol, Any}())
 
 setsolver!(r::StrategyRecord, s::Type{<:AbstractSolver}) = r.cur_solver = s
 getsolver(r::StrategyRecord) = r.cur_solver
+
+abstract type AbstractConquerStrategy <: AbstractStrategy end
+abstract type AbstractDivideStrategy <: AbstractStrategy end
+abstract type AbstractTreeSearchStrategy <: AbstractStrategy end
+
+"""
+    apply!(S::Type{<:AbstractStrategy}, args...)
```
guimarqu: Methods apply! must have the same arguments for all conquer and divide strategies because it is called by Coluna. The arguments should be indicated in the doc. I think we can "copy paste" the docstrings in `src/strategies/conquer/strategy.jl` and `src/strategies/divide/strategy.jl` and then delete these two files.

vitornesello: But for treesearch this is not the case. I will try to put all in the same file as you suggested.

PR #70, opened 2019-05-14
Cleaning: Problem setters + style

vitornesello:


src/formulation.jl
```diff
@@ -518,11 +518,16 @@ end
...
-    @logmsg LogLevel(0) string("Optimizing formulation ", getuid(form))
+    @logmsg LogLevel(-1) string("Optimizing formulation ", getuid(form))
+    @logmsg LogLevel(-3) "Coluna formulation before sync: "
+    @logmsg LogLevel(-3) form
     @logmsg LogLevel(-3) "MOI formulation before sync: "
     # _show_optimizer(form.moi_optimizer)
     sync_solver(form)
+    @logmsg LogLevel(-2) "Coluna formulation after sync: "
+    @logmsg LogLevel(-2) form
     @logmsg LogLevel(-2) "MOI formulation after sync: "
```
guimarqu: Something missing after this output

vitornesello: the line is commented because i cannot embed in the log_msg. we have to find a way to fix this

PR #72, opened 2019-05-15
Maximization bug fixes

vitornesello:
Fixing bugs related to the new maximisation feature

### DIFF
--- a/src/decomposition.jl
+++ b/src/decomposition.jl
@@ -186,6 +186,7 @@ function reformulate!(prob::Problem, annotations::Annotations,
     # Create master formulation
     master_form = Formulation{DwMaster}(
         prob.form_counter; parent_formulation = reformulation,
+        obj_sense = getobjsense(get_original_formulation(prob)),
         moi_optimizer = prob.master_factory()
     )
     setmaster!(reformulation, master_form)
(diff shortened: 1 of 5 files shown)


src/decomposition.jl
```diff
@@ -1,12 +1,13 @@
 set_loc_art_var(f::Formulation, constr_id::ConstrId) = setvar!(
-    f, string("local_art_", constr_id), MastArtVar; cost = 10.0,
+    f, string("local_art_", constr_id), MastArtVar;
+    cost = (getobjsense(f) == MinSense ? 1000.0 : -1000.0),
```
guimarqu: Did you intend to change the cost?

src/decomposition.jl
```diff
@@ -15,11 +16,15 @@ function initialize_local_art_vars(master::Formulation,
     for (constr_id, constr) in constrs_in_form
         v = setvar!(
             master, string("local_art_of_", getname(constr)),
-            MastArtVar; cost = 10000.0, lb = 0.0, ub = Inf,
-# cost = getincval(constr), lb = 0.0, ub = Inf,
-            kind = Continuous, sense = Positive
+            MastArtVar;
+            cost = (getobjsense(master) == MinSense ? 1000.0 : -1000.0),
```
guimarqu: cost ?

src/decomposition.jl
vitornesello: it was a mistake, actually function `set_loc_art_var` is not used`.

src/decomposition.jl
vitornesello: it was a mistake.

PR #74, opened 2019-05-15
Rework tree solver

vitornesello:


src/Coluna.jl
```diff
@@ -70,6 +71,7 @@ include("MOIinterface.jl")
 include("solvers/colgen.jl")
 include("solvers/masteripheur.jl")
 include("solvers/generatechildrennodes.jl")
+include("solvers/reformulationsolver.jl")
```
guimarqu: I think it should not be in folder solver since it is another type of solver

vitornesello: so should rename:
 - solvers/ -> nodesolvers/
and we can leave reformulationsolver.jl in src/ since it is the only one

PR #82, opened 2019-05-21
reduce log + show functions

vitornesello:


test/full_instances_tests.jl
```diff
@@ -0,0 +1,87 @@
...
+            master_factory = with_optimizer(GLPK.Optimizer),
+            pricing_factory = with_optimizer(GLPK.Optimizer)
+        )
+
+        problem, x, dec = CLD.GeneralizedAssignment.model_max(data, coluna)
+        JuMP.optimize!(problem)
+        @test abs(JuMP.objective_value(problem) - 580.0) <= 0.00001
+    end
+
+    # @testset "gap BIG instance" begin
```
guimarqu: Why do you keep this test if it is commented?

vitornesello: for us to test the big instance from time to time without deploying it

Issue #43, opened 2019-02-25
Pure phase 1 to Simplex-based colgen


vitornesello:
Implement pure phase 1 in order to handle correctly master infeasibility.

vitornesello:
Closed by #86

PR #86, opened 2019-05-21
ColumnGeneration phase one

vitornesello:
Fix issue #43


src/algorithms/colgen.jl
```diff
@@ -1,4 +1,4 @@
-struct ColumnGeneration <: AbstractAlgorithm end
+struct FullColumnGeneration <: AbstractAlgorithm end
```
guimarqu: why `Full`?

vitornesello: it does phase one and phase two. You dont like the name?

Issue #88, opened 2019-05-22
Handle global infeasibility


vitornesello:
After finding the best solution, should we run a checker agains the original formulation to determine if the solution is feasible or not?

vitornesello:
Closed by #90

PR #90, opened 2019-05-22
Introduce OptimizationResult{S}

### LINKED ISSUE #89: Homogeneous return structure for `optimize!` functions
This will make code cleaner and will be necessary to do nested

vitornesello:
Should fix #89 by introducing
```julia
mutable struct OptimizationResult{S<:AbstractObjSense}
    feasible::Bool # True iff feasible
    primal_bound::PrimalBound{S}
    dual_bound::PrimalBound{S}
    primal_sols::Vector{PrimalSolution{S}}
    dual_sols::Vector{DualSolution{S}}
end
```


vitornesello:
Now it also fixes #88

PR #99, opened 2019-05-24
Benders decomposition

guimarqu:


src/formulation.jl
```diff
@@ -74,6 +74,11 @@ getelem(f::Formulation, id::ConstrId) = getconstr(f, id)
 generatevarid(f::Formulation) = VarId(getnewuid(f.var_counter), f.uid)
 generateconstrid(f::Formulation) = ConstrId(getnewuid(f.constr_counter), f.uid)
 
+getmaster(f::Formulation{<:AbstractMasterDuty}) = f
```
vitornesello: The fact that `getmaster` of a master formulation returns itself is misleading. Do we really need this?

src/formulation.jl
```diff
@@ -142,8 +147,8 @@ function setvar!(f::Formulation,
                   is_active::Bool = true,
                   is_explicit::Bool = true,
                   moi_index::MoiVarIndex = MoiVarIndex(),
-                  members::Union{ConstrMembership,Nothing} = nothing)
-    id = generatevarid(f)
+                  members::Union{ConstrMembership,Nothing} = nothing,
+                  id = generatevarid(f))
```
vitornesello: Why would someone pass the `varid` when calling `setvar!`? This may cause nasty problems since the caller may pass an existing `id` (he does not know which ids were created so far).

src/filters.jl
```diff
@@ -9,6 +9,9 @@ _rep_of_orig_var_(v::Variable) = (getduty(v) isa OriginalRepresentatives)
 "Returns true if `id_v[2]` is a pricing subproblem variable and is currently active"
 _active_pricing_sp_var_(id_v::Pair{VarId,Variable}) = get_cur_is_active(id_v[2]) == true && getduty(id_v[2]) <: AbstractDwSpVar
 
+
```
vitornesello: extra blank line

src/clone.jl
```diff
@@ -10,7 +10,7 @@ function clone_in_formulation!(dest::Formulation,
         var_data = data
```
vitornesello: For file `clone.jl`: ok to remove all commented lines.

src/decomposition.jl
```diff
@@ -37,131 +37,119 @@ function initialize_global_art_vars(master::Formulation)
...
-    return vars, constrs
-end
-
-function build_dw_master!(prob::Problem,
-                          annotation_id::Int,
-                          reformulation::Reformulation,
-                          master_form::Formulation,
-                          vars_in_form::VarDict,
-                          constrs_in_form::ConstrDict,
+function build_benders_master!(prob::Problem,
```
vitornesello: weird indentation

src/decomposition.jl
```diff
@@ -170,65 +158,300 @@ function instanciatemaster!(prob::Problem, reform, ::Type{BD.Master}, ::Type{BD.
...
+function instantiatesp!(prob::Problem, reform, mast, ::Type{BD.BendersSepSp}, ::Type{BD.Benders})
+    form = Formulation{BendersSp}(
+        prob.form_counter; parent_formulation = mast,
+        obj_sense = getobjsense(mast)
+    )
+    add_benders_sep_sp!(reform, form)
+    return form
+end
+
+# Master of Dantzig-Wolfe decomposition
```
vitornesello: Can you add a comment to explain what functions `varexpduty` do?

vitornesello: one can deduce, but it is not evident.

src/decomposition.jl
```diff
@@ -170,65 +158,300 @@ function instanciatemaster!(prob::Problem, reform, ::Type{BD.Master}, ::Type{BD.
...
     end
-    createmaster!(form, prob, reform, ann, annotations, form_type, dec_type)
+    create_side_vars_constrs!(form)
+    create_artificial_vars!(form)
+    initialize_optimizer!(form, getoptbuilder(prob, ann))
     return
 end
 
-function registerformulations!(prob::Problem, annotations::Annotations, reform, 
+function buildformulations!(prob::Problem, annotations::Annotations, reform,
```
vitornesello: can you add the types in the signature?

src/decomposition.jl
```diff
@@ -170,65 +158,300 @@ function instanciatemaster!(prob::Problem, reform, ::Type{BD.Master}, ::Type{BD.
...
+    end
+    return
+end
+
+function create_side_vars_constrs!(sp::Formulation{BendersSp})
+    sp_coef = getcoefmatrix(sp)
+    sp_id = getuid(sp)
+    # Cost constraint
+    mast = getmaster(sp)
+    ν = setvar!(
```
vitornesello: `ν`: why this weird character?

vitornesello:
we should discuss whether we put the types in the function signature.

src/varconstr.jl
```diff
@@ -27,8 +27,11 @@ getmoirecord(vc::AbstractVarConstr) = vc.moirecord
 # Helpers for getters and setters that acces fields in a level
 # under Variable or Constraint
 
-getuid(vc::AbstractVarConstr) = getuid(vc.id)
-getform(vc::AbstractVarConstr) = getformuid(vc.id)
+getuid(vc::AbstractVarConstr) = getuid(getid(vc))
+getformuid(vc::AbstractVarConstr) = getformuid(getid(vc))
+getprocuid(vc::AbstractVarConstr) = getprocuid(getid(vc))
+getsortid(vc::AbstractVarConstr) = getsortid(getid(vc))
+
```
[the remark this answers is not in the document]
guimarqu: it's a bug of github

src/problem.jl
```diff
@@ -1,17 +1,55 @@
...
+    #     end
+    # end
+    # error("Cannot get parent annotation of $ann.")
+    for annotation in annotations.annotation_set
+        if BD.getformulation(annotation) == BD.Master
+            return annotation
+        end
+    end
+end
+
```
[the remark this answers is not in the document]
guimarqu: In BD we have the id of the parent annotation in each annotation. But it does not work because I didn't implemented it. (it's 0). We should move this method in BD before merging feature-benders in master

src/formulation.jl
guimarqu: Because I need to create a variable that have the same id of another (it's a clone of the id only). Id is an optional arg and the default value is creating a new id.
The user should not use this arg but we need it here.

src/decomposition.jl
guimarqu: it's the name of the variable in the draft and I think it's better than a long and incomprehensible abbreviation.

guimarqu: it's nu

vitornesello: what draft? `nu` or `v` is good enough no? I worry about people non being able to type this character. Also it looks very much like a `v`

src/formulation.jl
vitornesello: we change it we think of something cleaner.

PR #100, opened 2019-05-25
Define custom optimizer

vitornesello:


vitornesello:
Fisrt version that works, but need to improve user experience. Right now it is required for the user to know internal Coluna behaviours.

PR #126, opened 2019-06-10
function setrhs! that buffers the change

tbulhoes:


src/MOIinterface.jl
```diff
@@ -72,6 +72,20 @@ function update_constr_member_in_optimizer!(optimizer::MoiOptimizer,
     return
 end
 
+function update_constr_rhs_in_optimizer!(optimizer::MoiOptimizer, c::Constraint)
+    moi_c_index = getindex(getmoirecord(c))
+    rhs = getrhs(getcurdata(c))
```
vitornesello: There is a shortcut for that: `getcurrhs(c::Constraint)`

src/MOIinterface.jl
```diff
@@ -72,6 +72,20 @@ function update_constr_member_in_optimizer!(optimizer::MoiOptimizer,
...
+    sense = getcursense(c)
+    if sense == Greater
+        set(getinner(optimizer), ConstraintSet(), c, GreaterThan(rhs)) 
+    elseif sense == Less
+        set(getinner(optimizer), ConstraintSet(), c, LessThan(rhs)) 
+    else
+        set(getinner(optimizer), ConstraintSet(), c, EqualTo(rhs)) 
+    end
```
vitornesello: Use the function `get_moi_set` defined in types.jl
You need to put the `MOI.` prefix before entities defined in MOI, because it is an external packages and we did not call `using MathOptInterface`.
And you should pass `moi_c_index instead` of `c`
Try the following instead of the `if/else`:
```julia
MOI.set(getinner(optimizer), MOI.ConstraintSet(), moi_c_index, get_moi_set(sense))
```

vitornesello:
Can you merge this in your branch `feature-preprocessing` instead? I think these modifications are not being tested

PR #133, opened 2019-06-12
Uniform filters suggestion

### LINKED ISSUE #106: getvars(::Formulation) should return a membersvector
To avoid the need of both filters that take a tuple and the own object.
Currently it returns a VarDict

vitornesello:
Suggestion to fix #106


guimarqu:
It does not look uniform

PR #143, opened 2019-06-26
Draft version of Benders algorithm

FD-V:


src/decomposition.jl
```diff
@@ -46,280 +50,349 @@ function instantiatemaster!(prob::Problem, reform, ::Type{BD.Master}, ::Type{BD.
     return form
 end
 
-function instantiatemaster!(prob::Problem, reform, ::Type{BD.Master}, ::Type{BD.Benders})
-    form = Formulation{BendersMaster}(
+function instantiatemaster!(
+    prob::Problem, reform, ::Type{BD.Master}, ::Type{BD.Benders}
```
vitornesello: Can we have the argument type here? Should we do a PR just for these small things?

src/decomposition.jl
```diff
@@ -46,280 +50,349 @@ function instantiatemaster!(prob::Problem, reform, ::Type{BD.Master}, ::Type{BD.
...
+            lb = getperenelb(nu_var), ub = getpereneub(nu_var), 
+            kind = Continuous, sense = Free, is_explicit = true, 
+            id = getid(nu_var)
+        )                                 
     end
     return
 end
 
-function create_artificial_vars!(mast::Formulation{BendersMaster})
+function create_artificial_vars!(masterform::Formulation{BendersMaster})
```
vitornesello: Write this in the inline form?

guimarqu:
> I ignored the style. We should fix it little by little in when changing the code or by large batches in specific PR's

Yes, file by file while updating the doc

PR #144, opened 2019-06-26
Get multiplicity of subproblems from BD

### LINKED ISSUE #118: Bounds of convexity constraints in master
Hello to all.
While implementing the Capacitated Lot-sizing Problem with coluna I noticed that the convexity constraints are imposing a lb of 0 and a ub of 1. Thus, as for this problem I should have lb = ub = 1 in such constraints, the application does not work correctly, resulting in a solution value equal to zero.
I changed the rhs of the lower bound convexity constraint from 0 to 1 in decomposition.jl and it worked.
Having lb = ub = 1 in the convexity constraints is generic for column generation or coluna should have some way to set the sense of the convexity constraint?
Application is available in: [https://github.com/hhkramer/ColunaCodes/tree/master/clsp](url)
Best regards,
Hugo

guimarqu:
fix #118

### DIFF
--- a/src/decomposition.jl
+++ b/src/decomposition.jl
@@ -103,18 +103,20 @@ function instantiate_orig_constrs!(mast::Formulation{DwMaster}, orig_form::Formu
     return
 end
 
-function create_side_vars_constrs!(mast::Formulation{DwMaster}, orig_form::Formulation)
+function create_side_vars_constrs!(mast::Formulation{DwMaster}, orig_form::Formulation, annotations)
     coefmatrix = getcoefmatrix(mast)
     for sp in mast.parent_formulation.dw_pricing_subprs
         spuid = getuid(sp)
+        ann = get(annotations, sp)
         setupvars = filter(var -> getduty(var[2]) == DwSpSetupVar, getvars(sp))
         @assert length(setupvars) == 1
         setupvar = collect(values(setupvars))[1] # issue 106
         clonevar!(mast, setupvar, MasterRepPricingSetupVar, is_explicit = false)
         # create convexity constraint
         name = "sp_lb_$spuid"
+        lb_mult = Float64(BD.getminmultiplicity(ann))
         lb_conv_constr = setconstr!(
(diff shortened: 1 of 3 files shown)


src/problem.jl
```diff
@@ -4,6 +4,7 @@ mutable struct Annotations
     ann_per_constr::Dict{Id{Constraint}, BD.Annotation}
     vars_per_ann::Dict{BD.Annotation, Dict{Id{Variable},Variable}}
     constrs_per_ann::Dict{BD.Annotation, Dict{Id{Constraint},Constraint}}
+    form_per_ann::Dict{Int, BD.Annotation}
```
vitornesello: this name is weird, shouldn't it be `ann_per_form`?

vitornesello: or `ann_per_form_uid`

src/decomposition.jl
```diff
@@ -103,18 +103,20 @@ function instantiate_orig_constrs!(mast::Formulation{DwMaster}, orig_form::Formu
...
     coefmatrix = getcoefmatrix(mast)
     for sp in mast.parent_formulation.dw_pricing_subprs
         spuid = getuid(sp)
+        ann = get(annotations, sp)
         setupvars = filter(var -> getduty(var[2]) == DwSpSetupVar, getvars(sp))
         @assert length(setupvars) == 1
         setupvar = collect(values(setupvars))[1] # issue 106
         clonevar!(mast, setupvar, MasterRepPricingSetupVar, is_explicit = false)
         # create convexity constraint
         name = "sp_lb_$spuid"
```
vitornesello: can we also get rid of the temp var? and maybe do not use interpolation?

src/decomposition.jl
```diff
@@ -123,10 +125,9 @@ function create_side_vars_constrs!(mast::Formulation{DwMaster}, orig_form::Formu
         coefmatrix[getid(lb_conv_constr), getid(setupvar)] = 1.0
 
         name = "sp_ub_$spuid"
```
vitornesello: same for temp var

Issue #136, opened 2019-06-16
Wrong termination status


tbulhoes:
In the preprocessing tests, I am solving only the root node of a GAP instance and getting the termination status. If the problem is not solved in the root, termination status is INFEASIBLE even when  a primal solution has been found.

vitornesello:
fixed by #138 and #151

PR #151, opened 2019-06-27
Improve statuses and feasibility determination

vitornesello:


vitornesello:
should fix #136

src/algorithms/generatechildrennodes.jl
```diff
@@ -21,8 +21,8 @@ struct MostFractionalRule <: RuleForUsualBranching end
 function run!(::Type{GenerateChildrenNode}, formulation, node, strategy_rec, parameters)
     @logmsg LogLevel(0) "Run generate children nodes"
     algorithm_data =  GenerateChildrenNodeData(getincumbents(node), formulation)
-    if ip_gap(algorithm_data.incumbents) <= 0.0
-        @logmsg LogLevel(-1) string("Subtree is conquered, no need for branching.")
+    if !isfertile(node)
+        @logmsg LogLevel(1) "Node is not capable of generating childre, aborting branching"
```
guimarqu: typo

guimarqu: is not capable = cannot

src/problem.jl
```diff
@@ -68,7 +68,9 @@ function optimize!(prob::Problem, annotations::Annotations, params::Params)
     end
     println(_to)
     println("Terminated.")
-    @show getbestprimalsol(opt_result)
+    if nbprimalsols(opt_result) >= 1
+        @show getbestprimalsol(opt_result)
```
guimarqu: I don't think we should print the optimal solution. Jump provides a method so that the user can retrieve it.

src/solsandbounds.jl
```diff
@@ -131,6 +134,8 @@ function Base.isinteger(s::PrimalSolution)
     return true
 end
 
+isfractional(s::PrimalSolution) = !Base.isinteger(s)
```
guimarqu: AbstractSolution to make it available to DualSolution.

src/problem.jl
vitornesello: I will erase it when I merge your other PRs

PR #160, opened 2019-07-27
Reformulation solver rework

vitornesello:


src/algorithms/reformulationsolver.jl
```diff
@@ -118,11 +116,9 @@ function apply!(::Type{<:ReformulationSolver}, reform::Reformulation)
         && get_nb_treated_nodes(reform_solver) < _params_.max_num_nodes
     )
     determine_statuses(res, tree_fully_explored)
```
guimarqu: `!` missing at the end of `determine_statuses`. Otherwise, you can return `getresult(reform_solver)`.

vitornesello: I will open another PR to fix styleguide

PR #161, opened 2019-08-07
WIP : Gap functions

vitornesello:


src/solsandbounds.jl
```diff
@@ -46,28 +46,43 @@ isbetter(b1::DualBound{MinSense}, b2::DualBound{MinSense}) = b1.value > b2.value
...
+function relativegap(b1::PrimalBound{S}, b2::PrimalBound{S}) where {S}
+    b1.value == b2.value && return 0.0
+    return _relativegap(b1, b2)
+end
+
+function relativegap(b1::PrimalBound{S}, b2::DualBound{S}) where {S}
+    if b1.value == Inf || b2.value == Inf || b1.value == -Inf || b2.value == -Inf
+        return Inf
+    end
+    b1.value == b2.value && return 0.0
```
guimarqu: I usually put parenthesis around the statement. We should address this case in the style guide.

vitornesello: i did not put because there are only `||` operators. I agreewe shoudl discuss this.

guimarqu: I was talking about  `b1.value == b2.value && return 0.0`

PR #173, opened 2019-08-26
fix empty optimization result + test of infeasible ip restricted master

### LINKED ISSUE #167: error getbestprimasol when OptimizationResults empty
When the ip restricted master heuristic is applied on a master containing 0 cols, it should return no solution found.

But we have an error because of :
`getbestprimalsol(res::OptimizationResult) = res.primal_sols[1]`  (optimizationresults.jl:53)

guimarqu:
Fix #167

### DIFF
--- a/src/optimizationresults.jl
+++ b/src/optimizationresults.jl
@@ -50,8 +50,8 @@ getprimalsols(res::OptimizationResult) = res.primal_sols
 getdualsols(res::OptimizationResult) = res.dual_sols
 nbprimalsols(res::OptimizationResult) = length(res.primal_sols)
 nbdualsols(res::OptimizationResult) = length(res.dual_sols)
-getbestprimalsol(res::OptimizationResult) = res.primal_sols[1]
-getbestdualsol(res::OptimizationResult) = res.dual_sols[1]
+getbestprimalsol(res::OptimizationResult) = get(res.primal_sols, 1, nothing)
+getbestdualsol(res::OptimizationResult) = get(res.dual_sols, 1, nothing)
 setprimalbound!(res::OptimizationResult, b::PrimalBound) = res.primal_bound = b
 setdualbound!(res::OptimizationResult, b::DualBound) = res.dual_bound = b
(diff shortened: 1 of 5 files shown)


src/optimizationresults.jl
```diff
@@ -50,8 +50,8 @@ getprimalsols(res::OptimizationResult) = res.primal_sols
 getdualsols(res::OptimizationResult) = res.dual_sols
 nbprimalsols(res::OptimizationResult) = length(res.primal_sols)
 nbdualsols(res::OptimizationResult) = length(res.dual_sols)
-getbestprimalsol(res::OptimizationResult) = res.primal_sols[1]
-getbestdualsol(res::OptimizationResult) = res.dual_sols[1]
+getbestprimalsol(res::OptimizationResult) = get(res.primal_sols, 1, nothing)
```
vitornesello: Maybe we should leave it to the user to ask if there is a solution available in case he is not sure.
We could add a function `issolavailable`
Anyway after calling `getbestsol` he will need to check if the return value was `nothing`

test/unit/algorithms/masteripheur.jl
```diff
@@ -0,0 +1,32 @@
...
+            params = CL.Params(
+                global_strategy = CL.GlobalStrategy(InfeasibleMasterIpHeur, CL.NoBranching, CL.DepthFirst)
+            ),
+            default_optimizer = with_optimizer(GLPK.Optimizer)
+        )
+
+        problem, x, dec = CLD.GeneralizedAssignment.model(data, coluna)
+
+        JuMP.optimize!(problem)
+        @test JuMP.objective_value(problem) == Inf
```
vitornesello: Can you also check the feasibility status?

src/optimizationresults.jl
vitornesello: So what is the best behavior? In any case, the user should assure himself that there is at least one solution available, it he uses directly the result of the new `getbestsol` he might also get in trouble

vitornesello: It means we will have to add methods for functions that expect a solution all over the place.

PR #183, opened 2019-09-16
Uniformize filter functions

### LINKED ISSUE #106: getvars(::Formulation) should return a membersvector

guimarqu:
fix #106

### DIFF
--- a/src/containers/elements.jl
+++ b/src/containers/elements.jl
@@ -1,0 +1,36 @@
+struct ElemDict{VC <: AbstractVarConstr}
+    elements::Dict{Id{VC}, VC}
+end
+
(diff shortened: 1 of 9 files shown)


src/containers/elements.jl
```diff
@@ -0,0 +1,36 @@
+struct ElemDict{VC <: AbstractVarConstr}
+    elements::Dict{Id{VC}, VC}
+end
+
+function ElemDict{VC}() where {VC <: AbstractVarConstr}
```
vitornesello: inlined mode?

vitornesello: to all functions in this file actually

PR #186, opened 2019-09-20
Use Iterators.filter

### LINKED ISSUE #185: Using `filter` not a good idea?
Because `filter` creates a copy that contains the filtered value.

guimarqu:
Fix #185


src/containers/elements.jl
```diff
@@ -15,4 +15,8 @@ lastindex(d::ElemDict) = lastindex(d.elements)
 
 function Base.filter(f::Function, elems::ElemDict{VC}) where {VC}
```
vitornesello: do we need to keep this?

guimarqu: Yes I think

PR #187, opened 2019-09-24
Update Members Containers

guimarqu:


src/containers/members.jl
```diff
@@ -11,11 +11,10 @@ end
 Construct a `MembersVector` with indices of type `I`, elements of type `K`, and
 records of type `T`.
 
-The `MembersVector` maps each index to a tuple of element and record. The  
-use should use this structure like a `Dict{I,T}`. If the user looks for an index 
+The `MembersVector` maps each index to a tuple of element and record. This 
+structure must be use like a `Vector{I,T}`.  If the user looks for an index
```
vitornesello: extra space after the dot

src/containers/members.jl
```diff
@@ -235,11 +250,15 @@ function Base.getindex(m::MembersMatrix{I,K,J,L,T}, row_id::J, col_id::I) where
     end
 end
 
-function Base.getindex(m::MembersMatrix{I,K,J,L,T}, row_id::J, ::Colon) where {I,K,J,L,T}
+function Base.getindex(
+    m::MembersMatrix{I,K,J,L,T}, row_id::J, ::Colon
+) where {I,K,J,L,T}
     _getrecordvector!(m.rows, row_id, m.cols.elements, false)
 end
```
vitornesello: what about if we write like this?

```julia
Base.getindex(
    m::MembersMatrix{I,K,J,L,T}, row_id::J, ::Colon
) where {I,K,J,L,T}  = _getrecordvector!(
    m.rows, row_id, m.cols.elements, false
)
```

guimarqu: I don't like. I think the assignment form is good when the function fits on a line.

Issue #157, opened 2019-07-25
Numerical stability


vitornesello:
- [ ] Find places in the code where numerical instability is a problem
- [ ] Discuss, implement and write documentations
- [ ] Improve stability on gap functions
- [ ] Define integrality and optimiality tolerances

guimarqu:
I'll create a project for this one.

PR #193, opened 2019-09-30
Colgen algorithm style, params & tests

### LINKED ISSUE #162: Untested behavior if max nb of iterations reached in colgen
Solving the play gap instance with the simple branch and price strategy with nb_max_it = 10 in colgen algorithm yields an error.

guimarqu:
Fix #162


src/algorithms/colgen.jl
```diff
@@ -1,33 +1,61 @@
 Base.@kwdef struct ColumnGeneration <: AbstractAlgorithm
-    option::Bool = false
+    max_nb_iterations::Int = 1000
+    optimality_tol::Float64 = 1e-5
```
vitornesello: why is this specific to colgen?

guimarqu: no, but `optimality_tol` and `feasibility_tol` are also in BendersCutGen. We can let them here while #157 is open.

Issue #172, opened 2019-08-26
The error when I add Coluna in Julia 1.2


FuliXiong:
(v1.2) pkg> add https://github.com/atoptima/Coluna.jl.git
   Cloning git-repo `https://github.com/atoptima/Coluna.jl.git`
  Updating git-repo `https://github.com/atoptima/Coluna.jl.git`
  Updating git-repo `https://github.com/atoptima/Coluna.jl.git`
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package ColunaDemos [a54e61d4]:
 ColunaDemos [a54e61d4] log:
 ├─ColunaDemos [a54e61d4] has no known versions!
 └─restricted to versions * by Coluna [88b4ec78] — no versions left
   └─Coluna [88b4ec78] log:
     ├─possible versions are: 0.1.0 or uninstalled
     └─Coluna [88b4ec78] is fixed to version 0.1.0

PR #196, opened 2019-10-01
Clean deps

guimarqu:
Should fix installation bugs (see #172 or https://discourse.julialang.org/t/issues-install-coluna-jl-oj-julia-1-1/26298)


guimarqu:
Waiting for BlockDecomposition v1.0.1

PR #213, opened 2019-12-08
Nested enumeration container

guimarqu:
Structure to create a tree of duty using `UInt` instead of types. It'll allow us to have `VarData` and `ConstrData` bitstype.

Syntax to create a tree : 
```julia
Coluna.@nestedenum VarConstrDuty,
    A ← VarConstrDuty,
        A1 ← A,
        A2 ← A,
        A3 ← A,
    B ← VarConstrDuty,
        B1 ← B,
        B2 ← B,
        B3 ← B,
            B3A ← B3,
            B3B ← B3,
            B3C ← B3,
```
Syntax to know if duty inherits from another : 

```julia
A1 ← A # Return true
A1 ← B # Return false
!(A1 ← B) # Return true
```
@vitornesello your opinion about the syntax ?

### DIFF
--- a/src/containers/nestedenum.jl
+++ b/src/containers/nestedenum.jl
@@ -1,10 +1,10 @@
 abstract type NestedEnum end
 
 import Base.<=
+
 function <=(a::T, b::T) where {T <: NestedEnum}
     return a.id % b.id == 0
 end
-
 
 function _store!(expr::Symbol, i, names, parent_pos, leaves, primes)
     names[i] = expr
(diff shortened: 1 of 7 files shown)


vitornesello:
I don't like to much the arrows. Can we change it to something easy to find in the keyboard? like `<=` or `=>`. Also to keep some sort of consistency with dicts

`begin` and `end` are missing

guimarqu:
> I don't like too much the arrows. Can we change it to something easy to find in the keyboard? like `<=` or `=>`. Also to keep some sort of consistency with dicts
> 
> `begin` and `end` are missing

Ok for `<=`.  My first idea was to use `<:` but it's not possible to overload this operator.
It works without `begin` and `end`.

vitornesello:
so `begin` and `end` are optional?

guimarqu:
> so `begin` and `end` are optional?

Yes because I use commas. Maybe it's better without commas and using a block expression ?

vitornesello:
yeah, I prefer the block. It is also more common I think

src/containers/nestedenum.jl
```diff
@@ -0,0 +1,89 @@
...
+    length(r) == 0 && error("Unknow parent $(parent_name).")
+    length(r) > 1 && error("$parent_name registered more than once.")
+    parent_pos[i] = r[1]
+    names[i] = name
+    leaves[i] = true
+    primes[i] = Primes.prime(i)
+    return
+end
+
+function _compute_values!(values, parent_pos, primes)
```
vitornesello: Can you add a small comment to this function?

src/containers/nestedenum.jl
```diff
@@ -0,0 +1,89 @@
...
+    for i in 1:length(parent_pos)
+        factor = 1.0
```
vitornesello: Why is it not an `Int`?

vitornesello:
why do you need `leaves`?

PR #218, opened 2019-12-12
ElemDict & Members in Containers

guimarqu:


src/containers/members.jl
```diff
@@ -126,22 +126,24 @@ function Base.copy(vec::V) where {V <: MembersVector}
...
-        print(io, " ", id, " => (", getname(getelement(vec, id)), ", " , val, ")  ")
+        print(io, " ", id, " => " , val, " ")
     end
     print(io, "]")
 end
 
 struct MembersMatrix{I,K,J,L,T} <: AbstractMembersContainer
-    cols::MembersVector{I,K,MembersVector{J,L,T}}
-    rows::MembersVector{J,L,MembersVector{I,K,T}}
+    #matrix_csc::DynamicSparseArrays.MappedPackedCSC{}
```
vitornesello: why is it commented?

guimarqu: I forgot to delete... it doesn't matter because we'll delete this data structure.

PR #221, opened 2019-12-14
macro to export items of nestedenum

guimarqu:


guimarqu:
> Can't you create the nested enum and then export it? Do we really need to have both macros?

We need both macros because `@nestedenum` returns something like : 
```julia
struct Root <: NestedEnum value::Uint32 end
const ChildA = Root(2)
const ChildB = Root(3)
# etc
```

then we have to export all items

PR #225, opened 2019-12-18
Create submodule Algorithm

guimarqu:


src/Algorithm/interface.jl
```diff
@@ -1,3 +1,22 @@
+abstract type AbstractNode end
```
vitornesello: Why is node here?

PR #228, opened 2019-12-30
Var costs, lbs, and ubs stored in the formulation

guimarqu:
Adding package `DynamicSparseArrays` in Coluna.
Current cost, lb, and ub of variables are not stored in the Variable anymore but in the formulation.


src/MathProg/variable.jl
```diff
@@ -14,38 +11,66 @@ mutable struct VarData <: AbstractVcData
...
-getcost(v::VarData) = v.cost
-getlb(v::VarData) = v.lb
-getub(v::VarData) = v.ub
-
-setcost!(v::VarData, cost::Float64) = v.cost = cost
-setlb!(v::VarData, lb::Float64) = v.lb = lb
-setub!(v::VarData, ub::Float64) = v.ub = ub
-
-function setkind!(v::VarData, kind::VarKind)
+function _setkind!(v::VarData, kind::VarKind)
```
vitornesello: What about continuous variables?

src/MathProg/variable.jl
```diff
@@ -14,38 +11,66 @@ mutable struct VarData <: AbstractVcData
...
     if kind == Binary
         v.kind = Binary
         (v.lb < 0) && setlb!(v, 0.0)
         (v.ub > 1) && setub!(v, 1.0)
-    elseif kind == Integ
-        v.kind = Integ
     end
     return
 end
```
vitornesello: What about continuous variables?

src/MathProg/variable.jl
```diff
@@ -14,38 +11,66 @@ mutable struct VarData <: AbstractVcData
...
+    lb::Float64 = 0.0,
+    ub::Float64 = Inf,
+    kind::VarKind = Continuous,
+    sense::VarSense = Positive,
+    inc_val::Float64 = -1.0,
+    is_active::Bool = true,
+    is_explicit::Bool = true
+)
+    vc = VarData(cost, lb, ub, kind, sense, inc_val, is_active, is_explicit)
+    _setkind!(vc, kind) # should move in preprocessing ?
```
vitornesello: I think it is ok to leave it here because we might need this function after the after/outside preprocessing

src/MathProg/MOIinterface.jl
```diff
@@ -25,8 +25,9 @@ function compute_moi_terms(members::VarMembership)
     ]
 end
 
-function update_bounds_in_optimizer!(optimizer::MoiOptimizer,
+function update_bounds_in_optimizer!(form::Formulation,
```
vitornesello: not follwoing the style  :
```
longfunc(
    params
)
```

PR #231, opened 2020-01-02
Hide VarConstrData

guimarqu:
Remove the access to `cur_data` and `perene_data` because some data will be store elsewhere (see #228)

I create a `VarState` and a `ConstrState` to replace the use of `VarConstrData` when we prepare the formulation for a new node.


test/unit/variable.jl
```diff
@@ -2,37 +2,38 @@ function variable_unit_tests()
     var_data_getters_and_setters_tests()
     moi_var_record_getters_and_setters_tests()
     variable_getters_and_setters_tests()
+    return
```
vitornesello: sage

test/unit/variable.jl
```diff
@@ -2,37 +2,38 @@ function variable_unit_tests()
...
+
+    # ClF.setkind!(v_data, ClF.Binary)
+    # @test ClF.getkind(v_data) == ClF.Binary
+    # @test ClF.getlb(v_data) == 0.0
+    # @test ClF.getub(v_data) == -113.0
+
+    # ClF.setkind!(v_data, ClF.Integ)
+    # @test ClF.getkind(v_data) == ClF.Integ
+    # @test ClF.getlb(v_data) == 0.0
+    # @test ClF.getub(v_data) == -113.0
```
vitornesello: Why commented?

guimarqu: Because I'll change the test in a future PR

PR #233, opened 2020-01-10
get/set sense, kind & inc_val

guimarqu:
Delete `setpereneincval!`.

Rework methods

Rename `getsense`, `getkind` methods that translate the MOI struct to a Coluna struct.


src/MOIwrapper.jl
```diff
@@ -82,15 +82,15 @@ function get_var_kinds_and_bounds(src::MOI.ModelLike)
                 set = MOI.get(src, MOI.ConstraintSet(), moi_index)
                 id = func.variable.value
                 if S in [MOI.ZeroOne, MOI.Integer]
-                    kinds[id] = getkind(set)
+                    kinds[id] = tr_kind_MOI_to_Coluna(set)
```
vitornesello: what about `tr_MOI_kind_to_coluna` or `transform_MOI_kind_to_coluna`?

guimarqu: 1st one

PR #236, opened 2020-01-22
Duty{<:AbstractVarConstr} to move it in Id{<:AbstractVarConstr}

guimarqu:


src/Containers/nestedenum.jl
```diff
@@ -14,6 +14,13 @@ end
 
 # Store the item defined in expr at position i
 function _store!(expr::Expr, i, names, parent_pos, depths)
+    if i == 1 # parent can be a curly expression e.g. Duty{Variable}
+        expr.head == :curly || error("Syntax error : parent can be a Symbol or a curly expression.")
```
vitornesello: So if parent is not a curly expression you throw an error?

guimarqu: No I store the name which is a Symbol ( the _store! method with expr::Symbol )

PR #241, opened 2020-01-24
Dynamic Sparse Matrix for coefficient matrix

### LINKED ISSUE #198: Dynamic Sparse Matrix
Coefficient sparse matrix is stored using List of List format. This format has very bad performance for SpMV.
In the mid-term, we can try to improve LoL. 
In the long-term, we should try a better format like Dynamic CSR, Packed CSR or Hornet.

guimarqu:
Fix #198


src/Containers/members.jl
```diff
@@ -139,34 +139,60 @@ function Base.show(io::IO, vec::MembersVector{I,J,K}) where {I,J,K}
...
+    return MembersMatrix{I,J,T}(
+        DynamicSparseArrays.MappedPackedCSC(I,J,T),
+        DynamicSparseArrays.MappedPackedCSC(J,I,T)
+    )
+end
+
+function Base.setindex!(m::MembersMatrix, val, row_id, col_id)
+    m.cols_major[row_id, col_id] = val
+    m.rows_major[col_id, row_id] = val
+    m
```
vitornesello: no `return` keyword?

src/MathProg/optimizerwrappers.jl
```diff
@@ -108,15 +108,10 @@ function sync_solver!(optimizer::MoiOptimizer, f::Formulation)
     end
 
     # Add constrs
-    for id in buffer.constr_buffer.added
-        c = getconstr(f, id)
-        @logmsg LogLevel(-4) string("Adding constraint ", getname(f, c))
-        add_to_optimizer!(
-            f, c, filter(
-                vc -> getcurisactive(f, vc) && getcurisexplicit(f, vc),
```
vitornesello: why this filter is removed?

guimarqu: Because with the new matrix, `matrix[id, :]` creates a new container, and `filter` another one. So I put the condition in the method `add_to_optimizer!`.

PR #243, opened 2020-01-26
Speed up reduced costs computation

guimarqu:
3x ~ 9x faster on full instance tests


guimarqu:
Waiting for #247

guimarqu:
Closed by  #309

PR #247, opened 2020-01-28
Dynamic sparse vector for solutions

guimarqu:
- Replace the SortedDict by a PackedMemoryArray in the Solution container
- Fix a tolerance problem in `fill_primal_solution`.

### DIFF
--- a/src/parameters.jl
+++ b/src/parameters.jl
@@ -4,6 +4,7 @@ Base.@kwdef mutable struct Params
     integrality_tolerance::Float64 = 1e-5
     absolute_optimality_tolerance::Float64 = 1e-5
     relative_optimality_tolerance::Float64 = 1e-5
+    integrality_tol_digits::Int = 6 # because round(val, digits = 6)
     cut_up::Float64 = Inf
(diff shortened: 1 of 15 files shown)


src/parameters.jl
```diff
@@ -4,6 +4,8 @@ Base.@kwdef mutable struct Params
     integrality_tolerance::Float64 = 1e-5
     absolute_optimality_tolerance::Float64 = 1e-5
     relative_optimality_tolerance::Float64 = 1e-5
+    ϵ_tol::Float64 = 1e-6 # if - ϵ_tol < val < ϵ_tol, we consider val = 0
```
vitornesello: Maybe remove all other tolerances and leave this one as `tolerance`?

src/parameters.jl
```diff
@@ -4,6 +4,8 @@ Base.@kwdef mutable struct Params
...
+    integrality_tol_digits::Int = 6 # because round(val, digits = 6)
```
vitornesello: maybe `tolerance_digits`?

PR #272, opened 2020-02-27
Dual sol pool matrix is dynamic sparse

guimarqu:


src/MathProg/formulation.jl
```diff
@@ -200,46 +200,33 @@ function setdualsol!(
     new_dual_sol::DualSolution{S}
 )::Tuple{Bool,ConstrId} where {S<:Coluna.AbstractSense}
     ### check if dualsol exists  take place here along the coeff update
-
-    prev_dual_sols = getdualsolmatrix(form)
-
-    for (prev_dual_sol_id, prev_dual_sol) in columns(prev_dual_sols)
-        rhs = getdualsolrhss(form)[prev_dual_sol_id]
+    dual_sols = getdualsolmatrix(form)
+    dual_sol_rhss = getdualsolrhss(form)
```
vitornesello: `_rhss` ?

guimarqu: Because there are many of them

PR #277, opened 2020-03-04
Pricing oracles

### LINKED ISSUE #83: Improve ad-hoc pricing solver user experience
- [x] As a pricing callback

### LINKED ISSUE #206: Mapping Coluna Vars/Constrs <-> JuMP Vars/Constrs
Needed for :
- #83 
- defining branching strategies
- cuts callback interface

guimarqu:
Fix #83 
Fix #206 : map is already there for variables (field `varmap` of `Optimizer`)

### DIFF
--- a/src/MOIcallbacks.jl
+++ b/src/MOIcallbacks.jl
@@ -1,23 +1,61 @@
 function MOI.submit(
     model::Optimizer,
     cb::BD.PricingSolution{MP.OracleData},
+    cost::Float64,
     variables::Vector{MOI.VariableIndex},
     values::Vector{Float64}
 )
-    @show variables
-    @show values
+    form = cb.oracle_data.form 
+    S = getobjsense(form)
+    result = OptimizationResult{S}()
+    pb = PrimalBound(form, cost)
+
+    colunavarids = [_get_orig_varid_in_form(model, form, v) for v in variables]
+
+    # setup variable
+    setup_var_id = [id for (id,v) in Iterators.filter(
+        v -> (iscuractive(form, v.first) && iscurexplicit(form, v.first) && getduty(v.first) <= DwSpSetupVar),
+        getvars(form)
+    )][1]
+    push!(colunavarids, setup_var_id)
+    push!(values, 1.0)
+    pb += getcurcost(form, setup_var_id)
+
+    result.primal_bound = pb
+    push!(result.primal_sols, PrimalSolution(form, colunavarids, values, pb))
+    setfeasibilitystatus!(result, FEASIBLE)
+    setterminationstatus!(result, OPTIMAL)
+    cb.oracle_data.result = result
+    return
 end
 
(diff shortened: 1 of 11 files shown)


src/MOIcallbacks.jl
```diff
@@ -0,0 +1,61 @@
...
+)
+    form = cb.callback_data.form 
+    S = getobjsense(form)
+    result = OptimizationResult{S}()
+    pb = PrimalBound(form, cost)
+
+    colunavarids = [_get_orig_varid_in_form(model, form, v) for v in variables]
+
+    # setup variable
+    setup_var_id = [id for (id,v) in Iterators.filter(
```
vitornesello: isn't it very slow? to filter all the variables to get only one of them?

src/MOIcallbacks.jl
```diff
@@ -0,0 +1,61 @@
...
+    uid = getuid(callback_data.form)
+    axis_index_value = model.annotations.ann_per_form[uid].axis_index_value
+    return axis_index_value
+end
+
+function MOI.get(
+    model::Optimizer, vc::BD.PricingVariableCost{MP.PricingCallbackData}, 
+    x::MOI.VariableIndex
+)
+    form = vc.callback_data.form
```
vitornesello: why vc here? shouldnt it be var?

src/MOIcallbacks.jl
```diff
@@ -0,0 +1,61 @@
...
+
+function MOI.get(model::Optimizer, spid::BD.PricingSubproblemId{MP.PricingCallbackData})
+    callback_data = spid.callback_data
+    uid = getuid(callback_data.form)
+    axis_index_value = model.annotations.ann_per_form[uid].axis_index_value
+    return axis_index_value
+end
+
+function MOI.get(
+    model::Optimizer, vc::BD.PricingVariableCost{MP.PricingCallbackData},
```
vitornesello: why do we need `BD.PricingVariableCost{MP.PricingCallbackData}`. cant it be only `BD.VariableCost{MP.PricingCallbackData}`?

vitornesello: or `BD.VariablePrimal{MP.PricingCallbackData}`

src/MOIcallbacks.jl
guimarqu: We have nothing to select only this variable. Filtering all variables is the only way to get it for now.

src/MOIcallbacks.jl
guimarqu: I'll rename `pvc`

src/MOIcallbacks.jl
guimarqu: we can change the name but it will be for a future release of BlockDecomposition.

guimarqu: I'll try with `MOI.VariablePrimal`

src/MOIcallbacks.jl
vitornesello: ok. just something to keep in mind

PR #286, opened 2020-03-07
New structures for Solutions & Results

guimarqu:
Solution -> contains a formulation ref now

MoiResult -> to retrieve the result after solving a model with a MoiOptimizer/UserOptimizer

OptimizationOutput -> output of an algorithm & contains an OptimizationState

ObjValues -> contains incumbent bounds

OptimizationState : 

```julia
    OptimizationState(
        form; feasibility_status = UNKNOWN_FEASIBILITY, termination_status = NOT_YET_DETERMINED,
        ip_primal_bound = nothing, ip_dual_bound = nothing, lp_primal_bound = nothing, lp_dual_bound = nothing,
        max_length_ip_primal_sols = 1, max_length_lp_dual_sols = 1, max_length_lp_dual_sols = 1,
        insert_function_ip_primal_sols = bestbound, insert_function_lp_primal_sols = bestbound, 
        insert_function_lp_dual_sols = bestbound
        )
```

A convenient structure to maintain and return solutions and bounds of a formulation `form` during an optimization process. The feasibility and termination statuses are considered as unknown by default. You can define the initial incumbent bounds using `ip_primal_bound`, `ip_dual_bound`, `lp_primal_bound`, and `lp_primal_bound` keyword arguments. Incumbent bounds are set to infinite (according to formulation objective sense) by default. You can store three types of solutions `ip_primal_sols`, `lp_primal_sols`, and `lp_dual_sols`. These solutions are stored in three lists. Keywords `max_length_ip_primal_sols`, `max_length_lp_primal_sols`, and `max_length_lp_dual_sols` let you define the maximum size of the lists. Keywords `insert_function_ip_primal_sols`, `insert_function_lp_primal_sols`, and `insert_function_lp_dual_sols` let you provide a function to define the way you want to insert a new solution in each list. By default, lists are sorted by best bound.


src/Algorithm/interface.jl
```diff
@@ -54,77 +54,50 @@ getslavealgorithms!(
...
-function OptimizationOutput(incumb::Incumbents)
-    sense = getsense(incumb)
-    return OptimizationOutput{sense}(
-        OptimizationResult{sense}(
-            NOT_YET_DETERMINED, UNKNOWN_FEASIBILITY, get_ip_primal_bound(incumb),
-            get_ip_dual_bound(incumb), [], []
-        ), 
-        PrimalSolution{sense}(), DualBound{sense}()
-    )
+struct OptimizationOutput{F,S} <: AbstractOutput
```
vitornesello: Is this structure still going to be completed? Right now it is not consistent with the description in the PR.
Why do we need it?

PR #303, opened 2020-03-20
Update solveipform.jl

YP-Ye:
add run! method


guimarqu:
Can you please remove the method `run!` in `test/unit/algorithms/masteripheur.jl` ?

PR #308, opened 2020-04-14
Reorganisation of Algorithm submodule 

rrsadykov:
This takes into account the new OptimizationState structure

### DIFF
--- a/src/Algorithm/Algorithm.jl
+++ b/src/Algorithm/Algorithm.jl
@@ -25,7 +25,7 @@ import .MathProg: getfeasibilitystatus, getterminationstatus, setfeasibilitystat
     setterminationstatus!, isfeasible, get_ip_primal_bound, get_ip_dual_bound, 
     get_lp_primal_bound, get_lp_dual_bound, update_ip_primal_bound!, update_ip_dual_bound!, 
     update_lp_primal_bound!, update_lp_dual_bound!, set_ip_primal_bound!, 
-    set_ip_dual_bound!, set_lp_primal_bound!, set_lp_dual_bound!, ip_gap
+    set_ip_dual_bound!, set_lp_primal_bound!, set_lp_dual_bound!, ip_gap, lp_gap
 
 # Import to define run! method of algorithms
 import .ColunaBase: run!
@@ -37,18 +37,16 @@ include("utilities/record.jl")
 # Abstract algorithm
 include("interface.jl")
 
-# Here include slave algorithms used by conquer algorithms
+# Slave algorithms used by conquer algorithms
 include("solveipform.jl")
 include("solvelpform.jl")
 include("colgen.jl")
 include("benders.jl")
 include("preprocessing.jl")
 
-# Here include conquer algorithms
+# Algorithms and structures used by the tree search algorithm
+include("node.jl") 
 include("conquer.jl")
-
-include("node.jl") # TODO : break interdependance between node & Algorithm #224 & rm file
-
 include("divide.jl")
 
 # Here include divide algorithms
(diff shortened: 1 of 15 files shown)


rrsadykov:
Some questions: 
- why ObjValues is mutable? 
- in the src/optimize.jl::optimize!() function, probably, we can project the solution to the same OptimizationState without creating the new one?
- I did not well understand when we should copy solutions from one OptimizationState to another and when we should deepcopy? 

Remarks: 
- I decided to store OptimizationState in Node, this makes the code cleaner
- I decided to pass Node to ConquerAlgorithm, this also makes the code much cleaner.
- ConquerAlgorithm does not return anything now. It modifies the node optimization state directly in place, it also saves the record (future StorageState) to the node when needed
- We should probably do so profiling to see where is the time goes

guimarqu:
> Some questions:
> 
>     * why ObjValues is mutable?

Because you might want to update the value of an objective e.g. when you find a new feasible primal solution.

>     * in the src/optimize.jl::optimize!() function, probably, we can project the solution to the same OptimizationState without creating the new one?

Maybe yes

> 
>     * I did not well understand when we should copy solutions from one OptimizationState to another and when we should deepcopy?

You copy if you are sure you does not have another reference to the OptimizationState that can modify it. Otherwise, you deepcopy.

I think we deepcopy when we create a new node.

> 
> Remarks:
> 
>     * I decided to store OptimizationState in Node, this makes the code cleaner
> 
>     * I decided to pass Node to ConquerAlgorithm, this also makes the code much cleaner.
> 
>     * ConquerAlgorithm does not return anything now. It modifies the node optimization state directly in place, it also saves the record (future StorageState) to the node when needed

Ok, are you sure it works regardless the order in which the conquer algorithms are called ?

> 
>     * We should probably do so profiling to see where is the time goes

Otherwise ok.

rrsadykov:
> You copy if you are sure you does not have another reference to the OptimizationState that can modify it. Otherwise, you deepcopy.
> 
> I think we deepcopy when we create a new node.

We never modify solutions in OptimizationState, we can only add or replace solutions there. So we can always copy? 

Is this

`set_ip_primal_sol!(dest_state, get_best_ip_primal_sol(orig_state))`

is the same as

`set_ip_primal_sol!(dest_state, copy(get_best_ip_primal_sol(orig_state)))`?

> Ok, are you sure it works regardless the order in which the conquer algorithms are called ?

Conquer algorithm is called at most once for every node. If you want to run another conquer algorithm (as in strong branching), you first create another node and copy the node information there.

guimarqu:
> > You copy if you are sure you does not have another reference to the OptimizationState that can modify it. Otherwise, you deepcopy.
> > I think we deepcopy when we create a new node.
> 
> We never modify solutions in OptimizationState, we can only add or replace solutions there. So we can always copy?
> 
> Is this
> 
> `set_ip_primal_sol!(dest_state, get_best_ip_primal_sol(orig_state))`
> 
> is the same as
> 
> `set_ip_primal_sol!(dest_state, copy(get_best_ip_primal_sol(orig_state)))`?

It will not work because the copy method of PrimalSolution is not implemented. 
If we create a copy method like :  
```julia
copy(sol::Solution) = Solution(sol.model, sol.bound, sol.sol)
```
the solution and the copied solution share the same ref to the model, the bound and the sol. 
It depends on the definition of your method.

If you do a deepcopy, the solution and the deep copied solution share nothing.

I think copy is useless for Solution & we never modify solutions, so it's ok.

> 
> > Ok, are you sure it works regardless the order in which the conquer algorithms are called ?
> 
> Conquer algorithm is called at most once for every node. If you want to run another conquer algorithm (as in strong branching), you first create another node and copy the node information there.

ok

PR #322, opened 2020-04-26
Warm start for sub-solvers

guimarqu:


guimarqu:
@rrsadykov maybe I should rename the method `setprimalwarmstart!`

rrsadykov:
setbasis is good name for me. We also need getbasis.
Why do you use primal solution to set the basis? How does it work? 
In BaPCod Cplex interface, the basis contains variables and also constraints.  Do you think using MOI.VariablePrimalStart() is enough?

I will test getting/setting basis soon, I am now working on storing/restoring storages. I will make a draft pull request soon.

guimarqu:
Sorry I mixed two things. I will do a new PR for the basis. Here you give a primal solution to do a warm start.

rrsadykov:
What is the purpose of this warm start?

guimarqu:
> What is the purpose of this warm start?

Useless at the moment, so it stays here as a draft PR.
It may be more usefull for the IP solver

rrsadykov:
I think then it is better to move this to SolveIpForm. Actually the most useful for IP formulation  is to set the upper bound. I do not know whether IP solvers can make use of the primal solution. In principle, it may be useful only for heuristics embedded in IP solvers. 

For LP formulation, setting primal solution is not useful, we should set basis instead. Moreover, basis should not be set inside the algorithm, but in the node setup. I will implement this, I just need getbasis and setbasis function in the MathProg submodule.

rrsadykov:
I also think we should rename SolveIpForm to SolveMIPForm to make it visually more different from SolveLpForm.

guimarqu:
Stale

PR #324, opened 2020-04-27
get/set LP basis

guimarqu:
Doc about the BasisStatusCode in MOI
http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#MathOptInterface.BasisStatusCode

TODO :
- data structure to store/update the basis (and then setbasis! method)
- add support of set/get ConstraintBasisStatus in CPLEX
- add support of set ConstraintBasisStatus in GLPK
- exception handling


guimarqu:
Needed to enhance #344

guimarqu:
Stale

PR #335, opened 2020-05-11
silent mode for subsolvers

guimarqu:
use silent attribute to mute subsolvers instead of redirecting the output.

needed in #333


rrsadykov:
I think we need a silent option for the LP solver too (which is true by default). Sometimes you need to see the output of LP solver when you debug or when you have performance issues.

guimarqu:
> I think we need a silent option for the LP solver too (which is true by default). Sometimes you need to see the output of LP solver when you debug or when you have performance issues.

Yes, I put the option for both

PR #336, opened 2020-05-11
Create local artifical variables in setconstr!

guimarqu:
- keywork argument for `setconstr!` to create the local artificial variables with the constraint
- (de)activate a constraint and its artificial variables at the same time

### DIFF
--- a/src/MathProg/clone.jl
+++ b/src/MathProg/clone.jl
@@ -1,18 +1,20 @@
-function clonevar!(originform::Formulation,
-                   destform::Formulation,
-                   assignedform::Formulation,
-                   var::Variable,
-                   duty::Duty{Variable};
-                   name::String = getname(originform, var),
-                   cost::Float64 = getperenecost(originform, var),
-                   lb::Float64 = getperenelb(originform, var),
-                   ub::Float64 = getpereneub(originform, var),
-                   kind::VarKind = getperenekind(originform, var),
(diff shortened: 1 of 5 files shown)


rrsadykov:
`art_var_ids::Vector{VarId}` is needed only for master constraints. 
Is there a way to store duty-dependent data in constraints and variables?
For example, if a column is a route we would need to store a vector representing the sequence of arc ids. Non-linear constraints and cuts would need to store special information dependent on the constraint type in order to calculate coefficients for columns.

guimarqu:
> `art_var_ids::Vector{VarId}` is needed only for master constraints.
> Is there a way to store duty-dependent data in constraints and variables?
> For example, if a column is a route we would need to store a vector representing the sequence of arc ids. Non-linear constraints and cuts would need to store special information dependent on the constraint type in order to calculate coefficients for columns.

Ok but you cannot know that the column is a route. The duty is column not route. You should store that outside MathProg. Same for non-linear constraints.

rrsadykov:
Ok, for artificial variables we may leave like this for now. 
I agree that specific variable and constraint information should be stored outside MathProg. But MathProg should provide a way to associate this information to VarConstr. 
Later, we will need to implement management of dynamic variables and constraints: we should be able to delete a varconstr from memory permanently. And when a varconstr deleted, the associated information should be deleted too. We cannot determine that the associated information should be deleted outside MathProg.

PR #341, opened 2020-05-13
Seq time breakdown2

dianabarros:
Branch with modifications on Gurobi and tests.


guimarqu:
I cannot merge this branch in master because we cannot use Gurobi in the tests (otherwise we need a license). We will merge #333 but before I will replace Gurobi by a pricing callback. I'll try to do it this weekend.

I was saying that you should create a new branch where you fix the time limit bug only (change file `src/MathProg/optimizerwrappers.jl:171` you add the condition `terminationstatus != MOI.TIME_LIMIT`).
so then I can merge this new branch into master and the bug is fixed for everybody.

PR #345, opened 2020-05-20
Multi-threading on solving subproblems

dianabarros:
This implementation needs Gurobi to work.

### DIFF
--- a/src/Algorithm/colgen.jl
+++ b/src/Algorithm/colgen.jl
@@ -1,9 +1,9 @@
 Base.@kwdef struct ColumnGeneration <: AbstractOptimizationAlgorithm
     restr_master_solve_alg = SolveLpForm(get_dual_solution = true)
-    #TODO : pricing problem solver may be different depending on the 
+    #TODO : pricing problem solver may be different depending on the
     #       pricing subproblem
-    pricing_prob_solve_alg = SolveIpForm(deactivate_artificial_vars = false, 
-                                         enforce_integrality = false, 
+    pricing_prob_solve_alg = SolveIpForm(deactivate_artificial_vars = false,
+                                         enforce_integrality = false,
                                          log_level = 2)
     max_nb_iterations::Int = 1000
     optimality_tol::Float64 = 1e-5
@@ -25,7 +25,7 @@ end
 
 function get_storages_to_restore!(
     algo::ColumnGeneration, reform::Reformulation, storages_to_restore::StoragesToRestoreDict
-) 
+)
     master = getmaster(reform)
     add_storage!(storages_to_restore, master, MasterColumnsStorage, READ_AND_WRITE)
     get_storages_to_restore!(algo.restr_master_solve_alg, master, storages_to_restore)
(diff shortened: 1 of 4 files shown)


dianabarros:
This branch contains changes only related to the parallelization of the solve subproblems loop. For now, I’m using a global variable to say if we want to run the code using threads or not. I understand this might not be the best way of doing it, but I’d like to know what would be the best way of doing it for Coluna.
We could pass it as a parameter to the function but i don’t know at what point in the code we can set it.

I also didn't add Gurobi packages since a license is needed to run it, but it's important to note that this parallelization will only work with Gurobi. Should we keep it like this to merge with master?

src/Algorithm/colgen.jl
```diff
@@ -344,15 +343,31 @@ function solve_sps_to_gencols!(
...
 
     ### BEGIN LOOP TO BE PARALLELIZED
-    for (spuid, spdata) in spsdatas
-        gen_status, new_sp_sol_ids, sp_sol_ids_to_activate, sp_dual_contrib = solve_sp_to_gencol!(
-            algo, masterform, spdata, dual_sol, sp_lbs[spuid], sp_ubs[spuid]
-        )
-        if gen_status # else Sp is infeasible: contrib = Inf
-            recorded_sp_solution_ids[spuid] = new_sp_sol_ids
-            sp_solution_to_activate[spuid] = sp_sol_ids_to_activate
+    if Coluna.parallel
```
guimarqu, line 346: Instead of a global variable, it's better to create a new field in `ColumnGeneration` and set its default value to `false`. The name could be something like `solve_subproblems_parallel`. I'll create a new repo for preliminary experiments tomorrow.

guimarqu:
> I also didn't add Gurobi packages since a license is needed to run it, but it's important to note that this parallelization will only work with Gurobi. Should we keep it like this to merge with master?

If I have my own thread-safe algorithm to solve the subproblem, it should work, isn't it ?
If yes, I'll write an algorithm to solve the subproblem and put that in tests (but no in this PR, I'll do it later).

cristianabentes:
If I have my own thread-safe algorithm to solve the subproblem, it should work, isn't it ?
> Yes
If yes, I'll write an algorithm to solve the subproblem and put that in tests (but no in this PR, I'll do it later).
> Ok, but meanwhile we need Gurobi. We will leave the execution with Gurobi only in our branch. On the master branch the parallelization will be turned off until the home-made solution is ready.

src/Algorithm/colgen.jl
dianabarros: Fixed on c40fd87

src/Algorithm/colgen.jl
```diff
@@ -327,9 +326,10 @@ function updatereducedcosts!(reform::Reformulation, redcostsvec::ReducedCostsVec
 end
 
 function solve_sps_to_gencols!(
-    algo::ColumnGeneration, cgdata::ColGenRuntimeData, rfdata::ReformData, redcostsvec::ReducedCostsVector, 
+    algo::ColumnGeneration, cgdata::ColGenRuntimeData, rfdata::ReformData, redcostsvec::ReducedCostsVector,
     dual_sol::DualSolution, sp_lbs::Dict{FormId, Float64}, sp_ubs::Dict{FormId, Float64}
 )
+    threadstasks = Task[]
```
guimarqu, line 332: It looks unused

dianabarros: You're right, we don't need it anymore. I removed it.

guimarqu:
Close PR #341 #333 #217 #188

PR #351, opened 2020-05-25
Option to print branching tree in file

guimarqu:


rrsadykov:
@guimarqu I do not think it will work with strong branching. In this case, in `run_conquer_algorithm!()` it will quit before output to the dot file (here: `node.conquerwasrun && return`)

guimarqu:
> @guimarqu I do not think it will work with strong branching. In this case, in `run_conquer_algorithm!()` it will quit before output to the dot file (here: `node.conquerwasrun && return`)

Ok I'll fix it later this week & you'll review.

PR #352, opened 2020-05-25
fix most frac branching + cvrp demo for ci tests

### LINKED ISSUE #353: TreeSearchAlgorithm infinite branching loop when restricted LP master infeasible
**Describe the bug**
Title

**To Reproduce**
cvrp demo in https://github.com/atoptima/Coluna.jl/pull/352

**Expected behavior**
no infinite loop

guimarqu:
- add cvrp demo tests
- fix #353  (local art var in branching constraints + integer tol for Most frac rule)


src/Algorithm/branching/branchingalgo.jl
```diff
@@ -282,13 +283,22 @@ function run!(algo::StrongBranching, data::ReformData, input::DivideInput)::Divi
         end
 
         # sort branching candidates according to the selection criterion and remove excess ones
+        nb_candidates_kept = nb_candidates_needed
```
rrsadykov, line 286: I think it is much better not to generate branching groups with a very small distance to integer. In varbranching.jl, line 92, we check `if !isinteger(val)`. Instead, we should check if `val` is not within a tolerance to an integer.

guimarqu: Do I put a `int_tol` parameter in `VarBranchingRule` ?

rrsadykov: I think it is better to keep in StrongBranching and pass it to branching rules inside BranchingRuleInput

PR #358, opened 2020-06-01
add test for cvrp demo

guimarqu:


rrsadykov:
Actually, the demo is not correct, the lower bound becomes larger after cut separation than the optimal solution (784.0)

guimarqu:
> Actually, the demo is not correct, the lower bound becomes larger after cut separation than the optimal solution (784.0)

At the root node or deeper in the branching tree ?

rrsadykov:
Sorry, I messed up 748 and 784. The lower bound seems to be ok. However, it would be good to test a smaller instance and solve it to optimality. The root lower bound depends on the solver used. Also, if I put 785 as initial upper bound, optimal solution 784 is missed.

rrsadykov:
I made a smaller instances with 15 clients :
[A-n16-k3.vrp.zip](https://github.com/atoptima/Coluna.jl/files/4711844/A-n16-k3.vrp.zip)

Here is the optimal solution for it:
```
Best found solution of value 504 : 
Vehicle 1 of type 1 : depot -> 12 -> depot, demand = 21, time = 58
Vehicle 2 of type 1 : depot -> 6 -> 3 -> 2 -> 13 -> 7 -> 1 -> depot, demand = 90, time = 180
Vehicle 3 of type 1 : depot -> 5 -> 10 -> 15 -> 9 -> 11 -> 4 -> 8 -> 14 -> depot, demand = 95, time = 266

```

guimarqu:
> See the message above

It looks like we don't check if the solution returned by the LP restricted master is integer.
[cvrp.pdf](https://github.com/atoptima/Coluna.jl/files/4712064/cvrp.pdf)

guimarqu:
Coluna finds the optimal solution but we never test if the RM solution projected on original variables is integer.

```
  <it= 15> <et=36.63> <mst= 0.00> <sp= 0.01> <cols= 1> <DB=  494.0000> <mlp=  504.0000> <PB=504.0000>
  <it= 16> <et=36.65> <mst= 0.00> <sp= 0.01> <cols= 0> <DB=  504.0000> <mlp=  504.0000> <PB=504.0000>
[ Info: Dual bound reached primal bound.
	MC_2000059= 0.5
	MC_2000072= 1.0
	MC_2000118= 0.5
	MC_2000123= 0.5
	MC_2000148= 0.5
*******
	x[1,(1, 2)]= 1.0
	PricingSetupVar_sp_2= 3.0
	x[1,(1, 6)]= 1.0
	x[1,(1, 7)]= 1.0
	x[1,(1, 13)]= 2.0
	x[1,(1, 15)]= 1.0
	x[1,(2, 8)]= 1.0
	x[1,(3, 4)]= 1.0
	x[1,(3, 14)]= 1.0
	x[1,(4, 7)]= 1.0
	x[1,(5, 9)]= 1.0
	x[1,(5, 12)]= 1.0
	x[1,(6, 11)]= 1.0
	x[1,(8, 14)]= 1.0
	x[1,(9, 15)]= 1.0
	x[1,(10, 12)]= 1.0
	x[1,(10, 16)]= 1.0
	x[1,(11, 16)]= 1.0
Node is already conquered. No children will be generated
```

test/full_instances_tests.jl
```diff
@@ -1,12 +1,13 @@
 using LightGraphs
 
 function full_instances_tests()
-    generalized_assignment_tests()
-    capacitated_lot_sizing_tests()
-    lot_sizing_tests()
-    #facility_location_tests()
-    cutting_stock_tests()
+    # generalized_assignment_tests()
```
rrsadykov, line 4: Why you excluded other tests?

src/Algorithm/colgen.jl
```diff
@@ -591,6 +591,16 @@ function cg_main_loop!(algo::ColumnGeneration, cgdata::ColGenRuntimeData, rfdata
         if ip_gap(cg_optstate) < algo.optimality_tol
             setterminationstatus!(cg_optstate, OPTIMAL)
             @logmsg LogLevel(0) "Dual bound reached primal bound."
+
+            for (id, val) in get_best_lp_primal_sol(rm_optstate)
```
rrsadykov, line 595: This print should be removed?

rrsadykov:
I did not see any change to check the projected solution for integrality

guimarqu:
@rrsadykov It was a trash commit. Should we check integrality of the projected solution at each iteration ?

rrsadykov:
yes, in Bapcod we check on every iteration.

test/full_instances_tests.jl
```diff
@@ -178,7 +178,7 @@ function generalized_assignment_tests()
         problem, x, dec = CLD.GeneralizedAssignment.model(data, coluna)
 
         JuMP.optimize!(problem)
-        @test MOI.get(problem.moi_backend.optimizer, MOI.TerminationStatus()) == MOI.INFEASIBLE
+        @test_broken MOI.get(problem.moi_backend.optimizer, MOI.TerminationStatus()) == MOI.INFEASIBLE
```
guimarqu, line 181: Subproblem infeasibility is broken in column generation. I'll fix it in another PR.

guimarqu: It was broken before the merge of cut callback into master

src/Algorithm/colgen.jl
```diff
@@ -550,8 +550,14 @@ function cg_main_loop!(algo::ColumnGeneration, cgdata::ColGenRuntimeData, rfdata
         end
 
         if nb_lp_primal_sols(rm_optstate) > 0
-            set_lp_primal_sol!(cg_optstate, get_best_lp_primal_sol(rm_optstate))
+            rm_sol = get_best_lp_primal_sol(rm_optstate)
+            set_lp_primal_sol!(cg_optstate, rm_sol)
             set_lp_primal_bound!(cg_optstate, get_lp_primal_bound(rm_optstate))
+
+            proj_sol = proj_cols_on_rep(rm_sol, masterform)
+            if isinteger(proj_sol) && !contains(proj_sol, varid -> isanArtificialDuty(getduty(varid)))
```
rrsadykov, line 558: I think it is better to check for artificial variables in the solution before projecting the solution.

guimarqu: Indeed

PR #368, opened 2020-06-05
Finishing automatic smoothing

rrsadykov:


rrsadykov:
Stabilization is working fine. However, I am not happy with the performance. @guimarqu Guillaume, can you check which improvements are possible. For example functions `componentwisefunction()` and `update_alpha_automatically!()` in `colgenstabilization.jl` are slow. The last function which calculates the subgradient and updates alpha is about 100 (one hundred!) times slower than in BaPCod. I have an impression that iteration over solutions and membership matrices is very slow (and it allocates memory but it should not for me).

guimarqu:
Ok I’ll look tomorrow for performance. I can tell you that `matrix[id, :]` is inefficient for now because it creates a new array. I will implement `views` in DynamicSparseArrays to retrieve constraints and columns faster.

Project.toml
```diff
@@ -5,12 +5,17 @@ version = "0.3.0"
...
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 DynamicSparseArrays = "8086fd22-9a0c-46a5-a6c8-6e24676501fe"
+GLPK = "60bf3e95-4087-53dc-ae20-288a0d20c6a6"
+JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
+LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
guimarqu, line 18: new dependencies to remove

src/Algorithm/colgen.jl
```diff
@@ -298,16 +290,22 @@ function solve_sp_to_gencol!(
     update_pricing_target!(spform)
 
     # Solve sub-problem and insert generated columns in master
-    output = run!(algo.pricing_prob_solve_alg, spdata, OptimizationInput(OptimizationState(spform)))
+    TO.@timeit Coluna._to "Solve pricing problem" begin
```
guimarqu, line 293: TimerOutput not thread safe

src/Algorithm/colgenstabilization.jl
```diff
@@ -52,6 +55,7 @@ function componentwisefunction(in_dual_sol::DualSolution, out_dual_sol::DualSolu
     value::Float64 = 0.0
     out_next = iterate(out_dual_sol)
     in_next = iterate(in_dual_sol)
+    number = 0
```
guimarqu, line 58: what is it ?

src/Algorithm/colgen.jl
guimarqu: TimerOutput cannot be used in `solve_sp_to_gencol!` because it is not thread safe.

guimarqu:
> Ok, I have removed timers. But this is annoying. Is there a way to include timers only if non-parallel version is used (`solve_subproblems_parallel = false`)?

I don't know.

PR #380, opened 2020-06-29
Rewrite MOI wrapper

guimarqu:
- Implement automatic copy
- Move decomposition outside `MathProg`
- Move data structure that stores `Annotation`s outside `MathProg`
- Allow use of MOI direct mode

### DIFF
--- a/src/MOIwrapper.jl
+++ b/src/MOIwrapper.jl
@@ -284,7 +284,6 @@ function MOI.get(
 )
     orig_form = get_original_formulation(model.inner)
     lb = getperenlb(orig_form, model.vars[MOI.VariableIndex(index.value)])
-    ub = getperenub(orig_form, model.vars[MOI.VariableIndex(index.value)])
     return MOI.GreaterThan(lb)
 end
 
@@ -293,7 +292,6 @@ function MOI.get(
     index::MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}}
 )
     orig_form = get_original_formulation(model.inner)
-    lb = getperenlb(orig_form, model.vars[MOI.VariableIndex(index.value)])
     ub = getperenub(orig_form, model.vars[MOI.VariableIndex(index.value)])
     return MOI.LessThan(ub)
 end
@@ -348,7 +346,7 @@ function MOI.set(
     model::Coluna.Optimizer, ::MOI.VariableName, varid::MOI.VariableIndex, name::String
 )
     var = model.vars[varid]
-    # set perene name
+    # TODO : rm set perene name
     var.name = name
     return
 end
@@ -376,7 +374,7 @@ function MOI.set(
     model::Coluna.Optimizer, ::MOI.ConstraintName, constrid::MOI.ConstraintIndex, name::String
 )
     constr = model.constrs[constrid]
-    # set perene name
+    # TODO : rm set perene name
     constr.name = name
     return
 end
@@ -421,7 +419,7 @@ function MOI.set(
     for term in func.terms
         var = model.vars[term.variable_index]
         cost = term.coefficient
-        # set peren cost 
+        # TODO : rm set peren cost
         var.perendata.cost = cost
         var.curdata.cost = cost
     end
@@ -435,7 +433,7 @@ function MOI.set(
 )
     model.objective_type = SINGLE_VARIABLE
     var = model.vars[func.variable]
-    # set perene cost
+    # TODO : rm set perene cost
     var.perendata.cost = 1.0
     var.curdata.cost = 1.0
     return
--- a/src/MOIwrapper.jl
+++ b/src/MOIwrapper.jl
@@ -1,29 +1,42 @@
-const SupportedObjFunc = Union{MOI.ScalarAffineFunction{Float64},
-                               MOI.SingleVariable}
+const CleverDicts = MOI.Utilities.CleverDicts
 
-const SupportedVarSets = Union{MOI.ZeroOne,
-                               MOI.Integer,
-                               MOI.LessThan{Float64},
-                               MOI.EqualTo{Float64},
-                               MOI.GreaterThan{Float64}}
+const SupportedObjFunc = Union{MOI.ScalarAffineFunction{Float64},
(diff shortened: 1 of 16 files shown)


test/full_instances_tests.jl
```diff
@@ -100,6 +100,9 @@ function generalized_assignment_tests()
         )
 
         model, x, dec = CLD.GeneralizedAssignment.model(data, coluna)
+
+        println(model)
```
vitornesello, line 104: is this print intended?

guimarqu: Yes because it calls few MOI.get implemented in MOIwrapper.jl (we will have to do real tests next year)

src/MOIwrapper.jl
```diff
@@ -72,175 +85,407 @@ function MOI.optimize!(optimizer::Optimizer)
...
 end
 
-function MOI.supports(optimizer::Optimizer,
-        ::MOI.ObjectiveFunction{<: SupportedObjFunc})
-    return true
+############################################################################################
+# Add constraint
+############################################################################################
+function _constraint_on_variable!(var::Variable, ::MOI.Integer)
+    # set perene data
```
vitornesello, line 109: this comment seems misleading

src/MOIwrapper.jl
```diff
@@ -72,175 +85,407 @@ function MOI.optimize!(optimizer::Optimizer)
...
-function getvarcosts(src::MOI.ModelLike)
-    # We need to increment values of cost_rhs with += to handle cases like $x_1 + x_2 + x_1$
-    # This is safe becasue the variables are initialized with a 0.0 cost_rhs
-    costs = Dict{Int,Float64}()
-    obj = MOI.get(src, MoiObjective())
-    for term in obj.terms
-        id = term.variable_index.value
-        costs[id] = get(costs, id, 0.0) + term.coefficient
+function _constraint_on_variable!(var::Variable, ::MOI.ZeroOne)
+    # set perene data
```
vitornesello, line 116: this comment seems misleading

src/MOIwrapper.jl
```diff
@@ -72,175 +85,407 @@ function MOI.optimize!(optimizer::Optimizer)
...
-                        end
-                    end
-                end
-            end
+    constr = setconstr!(
+        orig_form, "c", OriginalConstr;
+        rhs = MathProg.convert_moi_rhs_to_coluna(set),
+        kind = Essential,
+        sense = MathProg.convert_moi_sense_to_coluna(set),
+        inc_val = 10.0,
```
vitornesello, line 181: why `inc_val = 10.0` Can we have a comment explaining?

guimarqu: I don't know why.

src/MOIwrapper.jl
guimarqu: We may have to implement setperencost!, setperenub!. I would like to think before doing it because I don't like the idea of setperen...

src/MOIwrapper.jl
```diff
@@ -72,175 +85,407 @@ function MOI.optimize!(optimizer::Optimizer)
...
+    return S(rhs)
+end
+
+function MOI.get(
+    model::Coluna.Optimizer, ::MOI.ConstraintSet,
+    index::MOI.ConstraintIndex{MOI.SingleVariable, MOI.GreaterThan{Float64}}
+)
+    orig_form = get_original_formulation(model.inner)
+    lb = getperenlb(orig_form, model.vars[MOI.VariableIndex(index.value)])
+    ub = getperenub(orig_form, model.vars[MOI.VariableIndex(index.value)])
```
vitornesello, line 287: why getting the `ub` here?

src/MOIwrapper.jl
```diff
@@ -72,175 +85,407 @@ function MOI.optimize!(optimizer::Optimizer)
...
+    return MOI.GreaterThan(lb)
+end
+
+function MOI.get(
+    model::Coluna.Optimizer, ::MOI.ConstraintSet,
+    index::MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}}
+)
+    orig_form = get_original_formulation(model.inner)
+    lb = getperenlb(orig_form, model.vars[MOI.VariableIndex(index.value)])
```
vitornesello, line 296: why getting the `lb` here?

src/MOIwrapper.jl
```diff
@@ -72,175 +85,407 @@ function MOI.optimize!(optimizer::Optimizer)
...
+    constr = model.constrs[constrid]
+    # set perene name
+    constr.name = name
+    return
+end
 
-    moi_uid_to_coluna_id = Dict{Int,VarId}()
-    create_origvars!(orig_form, dest, src, costs, kinds, lbs, ubs, copy_names, moi_uid_to_coluna_id)
-    create_origconstrs!(orig_form, dest, src, copy_names, moi_uid_to_coluna_id)
+function MOI.get(model::Coluna.Optimizer, ::MOI.ConstraintName, index::MOI.ConstraintIndex)
```
vitornesello, line 384: Not throwing `MOI.InvalidIndex` here?

src/MOIwrapper.jl
vitornesello: I think the comment is useless because it says exactly what is written in the following line.

Also, we are setting both perene and current data. So it is like the comment is there just to stress the fact that on top of the current data, the perene is also being set, which is an important thing because it is not modified anywhere else, right?

src/MOIwrapper.jl
guimarqu: done

src/MOIwrapper.jl
guimarqu: done

src/MOIwrapper.jl
guimarqu: I changed the comment to make it more explicit. In the future, I would like to avoid changing perene data

src/MOIwrapper.jl
guimarqu: I'll fix this by testing the MOI wrapper like CPLEX, Gurobi...

PR #385, opened 2020-08-04
Draft to get benchmarks

guimarqu:
Very dirty, do not merge.


guimarqu:
Close because done in #452

PR #387, opened 2020-09-14
Simplification of the interface for algorithms 

rrsadykov:
Before implementing diving, I have simplified the interface to algorithms. Now every algorithm should only communicate all its slave algorithms (if any) and all storages it uses with the usage mode (if any).
I have divided algorithms into manager algorithms and worker algorithms (see explanation in the beginning of file src/Algorithm/interface.jl).
I have also tried to simplify the code for storing/restoring storages.


guimarqu:
why tests don't pass ?

rrsadykov:
Sorry, I will correct

rrsadykov:
Corrected, now the tests should pass

src/Algorithm/interface.jl
```diff
@@ -1,7 +1,34 @@
+"""
+    About algorithms
+    ----------------
+
+    An algorithm is a procedure with a known interface (input and output) applied to a data.
+    An algorithm can use storages inside the data to keep its computed data between different
+    runs of the algorithm or between runs of different algorithms.
+    The algorithm itself contains only its parameters. 
+
+    Parameters of an algorithm may contain its slave algorithms which used by it. Therefore,
```
guimarqu, line 10: Since it's a tree of algorithms, I think "child algorithms" will be more explicit than "slave algorithms"

src/Algorithm/interface.jl
```diff
@@ -1,7 +1,34 @@
...
+    Algorithms are divided into two types : "manager algorithms" and "worker algorithms". 
+    Worker algorithms just continue the calculation. They do not store and restore storages 
+    as they suppose it is done by their master algorithms. Manager algorithms may divide 
+    the calculation flow into parts. Therefore, they store and restore storages to make sure 
+    that their slave worker algorithms have storages prepared. 
+    A worker algorithm cannot have slave manager algorithms. 
+
+    Examples of manager algorithms : TreeSearchAlgorithm (which covers both BCP algorithm and 
+    diving algorithm), conquer algorithms, strong branching, branching rule algorithms 
+    (which create child nodes). Examples of worker algorithms : column generation, SolveIpForm,
```
guimarqu, line 23: A worker algorithm is a leaf of the tree ? The name can be leaf algorithm ?

src/Algorithm/interface.jl
```diff
@@ -17,42 +44,37 @@ abstract type AbstractOutput end
...
 
+ismanager(algo::AbstractAlgorithm) = false
+
 """
-    run!(algo::AbstractAlgorithm, model::AbstractModel, input::AbstractInput)::AbstractOutput
+    get_slave_algorithms(::AbstractAlgorithm, ::AbstractModel)::Vector{Tuple{AbstractAlgorithm, AbstractModel}}
 
-    Runs the algorithm. The storage of the algorithm can be obtained by asking
-    the formulation. Returns algorithm's output.    
+    Every algorithm should communicate its slave algorithms and the model to which
```
guimarqu, line 55: should or must ?

guimarqu (commented):
I think we can improve names

- slave algorithms -> child algorithms

- worker algorithms -> leaf algorithms
-  managing algorithms -> algorithms that are not leaves  

?

src/Algorithm/interface.jl
rrsadykov: "slave" and "child" is the same for me, we can use "child" if you want
The problem is that a worker algorithm is not necessary a leaf algorithm. Col. gen. is a worker algorithm, but it has child algorithms.

src/Algorithm/interface.jl
rrsadykov: A worker is not necessarily a leaf.

guimarqu: Yes sorry I misunderstood. Ok for worker/manager

src/Algorithm/interface.jl
guimarqu: Yes I prefer parent/child because we talk about the position of the algo in the tree. Besides, I think "master/slave" may create confusion with "manager/worker".

rrsadykov:
Ok, I have renamed slave algorithms to child algorithms. I think we can merge now

guimarqu (approved):

rrsadykov:
I am not authorised to merge. Can you give me this authorisation?

guimarqu:
> I am not authorised to merge. Can you give me this authorisation?

and now ?

rrsadykov:
Ok, now I can. What should I do : squash and merge or rebase and merge?

guimarqu:
Always squash & merge

PR #394, opened 2020-10-06
Improve termination statuses

guimarqu:
I kept `NODE_LIMIT` because it can be usefull when we call subsolvers through MOI, or when we only solve the root node.

I replaced `NOT_YET_DETERMINED`  by  `UNKOWN_TERMINATION` (symmetric to `UNKNOWN_FEASIBILITY`) & `UNDEFINED`.

### DIFF
--- a/src/Algorithm/colgenstabilization.jl
+++ b/src/Algorithm/colgenstabilization.jl
@@ -105,7 +105,7 @@ function linear_combination(in_dual_sol::DualSolution, out_dual_sol::DualSolutio
     for (i, constrid) in enumerate(constrids)
         bound += constrvals[i] * getcurrhs(form, constrid) 
     end
-    return DualSolution(form, constrids, constrvals, bound)
+    return DualSolution(form, constrids, constrvals, bound, FEASIBLE_SOL)
 end
 
 function update_stab_after_rm_solve!(
@@ -144,7 +144,7 @@ function update_alpha_automatically!(
 
     # first we calculate the in-sep direction
     constrids, constrvals = componentwisefunction(smooth_dual_sol, storage.stabcenter, -)
-    in_sep_direction = DualSolution(master, constrids, constrvals, 0.0)
+    in_sep_direction = DualSolution(master, constrids, constrvals, 0.0, FEASIBLE_SOL)
     in_sep_dir_norm = norm(in_sep_direction)
 
     # we initialize the subgradient with the right-hand-side of all master constraints
@@ -158,7 +158,7 @@ function update_alpha_automatically!(
             push!(constrrhs, getcurrhs(master, constrid))
         end 
     end
-    subgradient = DualSolution(master, constrids, constrrhs, 0.0)
+    subgradient = DualSolution(master, constrids, constrrhs, 0.0, FEASIBLE_SOL)
     
     # we calculate the subgradient at the sep point
     for (constrid, value) in subgradient_contribution
--- a/src/Algorithm/colgenstabilization.jl
+++ b/src/Algorithm/colgenstabilization.jl
@@ -105,7 +105,7 @@ function linear_combination(in_dual_sol::DualSolution, out_dual_sol::DualSolutio
     for (i, constrid) in enumerate(constrids)
         bound += constrvals[i] * getcurrhs(form, constrid) 
     end
-    return DualSolution(form, constrids, constrvals, bound, FEASIBLE_SOL)
+    return DualSolution(form, constrids, constrvals, bound, UNKNOWN_FEASIBILITY)
 end
 
 function update_stab_after_rm_solve!(
@@ -144,7 +144,7 @@ function update_alpha_automatically!(
 
     # first we calculate the in-sep direction
     constrids, constrvals = componentwisefunction(smooth_dual_sol, storage.stabcenter, -)
-    in_sep_direction = DualSolution(master, constrids, constrvals, 0.0, FEASIBLE_SOL)
+    in_sep_direction = DualSolution(master, constrids, constrvals, 0.0, UNKNOWN_FEASIBILITY)
     in_sep_dir_norm = norm(in_sep_direction)
 
     # we initialize the subgradient with the right-hand-side of all master constraints
@@ -158,7 +158,7 @@ function update_alpha_automatically!(
             push!(constrrhs, getcurrhs(master, constrid))
         end 
     end
-    subgradient = DualSolution(master, constrids, constrrhs, 0.0, FEASIBLE_SOL)
+    subgradient = DualSolution(master, constrids, constrrhs, 0.0, UNKNOWN_FEASIBILITY)
     
     # we calculate the subgradient at the sep point
     for (constrid, value) in subgradient_contribution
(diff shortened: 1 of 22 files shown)


rrsadykov:
Why don't we use the termination status only? It can return also infeasible. 

I have checked : there is a single termination status in MOI, which may return INFEASIBLE. May be we just use `TerminationStatusCode` of MOI?

guimarqu:
> Why don't we use the termination status only? It can return also infeasible.
> 
> I have checked : there is a single termination status in MOI, which may return INFEASIBLE. May be we just use `TerminationStatusCode` of MOI?

We don't use MOI TerminationStatusCode because otherwise a package that contains an algorithm for Coluna will have MOI in its dependencies.

We separated Termination & Feasibility because Feasiblity can be also used for a solution (not only for the problem).

rrsadykov:
> We don't use MOI TerminationStatusCode because otherwise a package that contains an algorithm for Coluna will have MOI in its dependencies.

If a package depends on Coluna, it automatically depends on MOI, no? 
 
> We separated Termination & Feasibility because Feasibility can be also used for a solution (not only for the problem).

Nothing prevents us from using different statuses for algorithms and solutions. In MOI, they use different statuses. 
It is misleading for me to have two statuses. If feasibility status is infeasible, which value should the termination status take?

guimarqu:
> If a package depends on Coluna, it automatically depends on MOI, no?

Yes but this dependence won't be in the Project file but only in the Manifest. It's a "weaker" dependence.
 
> Nothing prevents us from using different statuses for algorithms and solutions. In MOI, they use different statuses.
> It is misleading for me to have two statuses for the algorithm. If feasibility status is infeasible, which value should the termination status take?

Yes I agree, we go for a `TerminationStatus` and `SolutionStatus` ?

`TerminationStatus` : `OPTIMAL`, `INFEASIBLE`, `TIME_LIMIT`, `NODE_LIMIT`, `OTHER_LIMIT`, `UNKNOWN_TERMINATION_STATUS`, `UNDEFINED_TERMINATION_STATUS`

`SolutionStatus` : `FEASIBLE_SOL`, `INFEASIBLE_SOL`, `EMPTY_SOLUTION`,  `UNKNOWN_SOLUTION_STATUS`, `UNDEFINED_SOLUTION_STATUS`

rrsadykov:
Ok for me, just one question: what is the difference between UNKNOWN_TERMINATION_STATUS and UNDEFINED_TERMINATION_STATUS?

guimarqu:
> Ok for me, just one question: what is the difference between UNKNOWN_TERMINATION_STATUS and UNDEFINED_TERMINATION_STATUS?

`UNKOWN` it's before the algorithm, `UNDEFINED` it's after the algorithm & it's used only when the subsolver is called through MOI and that MOI returns a status that we do not cover. Maybe we can call it `UNCOVERED`...

guimarqu:
I'll try to avoid the use of UNKNOWN. If we don't need it, I'll remove it.

rrsadykov:
Ok for me

rrsadykov:
I did not understand, why you set the solution infeasible status in the Benders algorithm, and termination infeasible status in for the column generation (in the case of infeasibility)?

guimarqu:
I have a doubt about statuses in Benders, I'll check & finish tomorrow.
I think it's ok for column generation.

Guillaume Marques

rrsadykov:
Ok. I also think that you should set the solution status when you change the solution in `OptimizationState`.

guimarqu:
> Ok. I also think that you should set the solution status when you change the solution in `OptimizationState`.

How do you know if the solution pushed into the `OptimizationState` is feasible or not ?

rrsadykov:
> > Ok. I also think that you should set the solution status when you change the solution in `OptimizationState`.
> 
> How do you know if the solution pushed into the `OptimizationState` is feasible or not ?

The algorithm which creates the solution should know this.

rrsadykov:
Also, there are many solutions in `OptimizationState`, the solution status refers to which one?

guimarqu:
Ok so I suggest to move the `SolutionStatus` into `Solution`.
It will mean that the solution is (in)feasible with regard to the model referenced.

Then, we'll have to give the status of the solution at its instantiation.

src/Algorithm/utilities/optimizationstate.jl
```diff
@@ -335,7 +336,7 @@ end
 function Base.print(io::IO, form::AbstractFormulation, optstate::OptimizationState)
     println(io, "┌ Optimization state ")
     println(io, "│ Termination status: ", optstate.termination_status)
-    println(io, "│ Feasibility status: ", optstate.feasibility_status)
+    println(io, "│ Solution status: ", optstate.solution_status)
```
rrsadykov, line 339: No more solution status for `OptimizationState`, no?

rrsadykov:
Otherwise, this merge is ok for me.

src/Algorithm/utilities/optimizationstate.jl
guimarqu: I'll fix it.

src/Algorithm/colgenstabilization.jl
```diff
@@ -105,7 +105,7 @@ function linear_combination(in_dual_sol::DualSolution, out_dual_sol::DualSolutio
     for (i, constrid) in enumerate(constrids)
         bound += constrvals[i] * getcurrhs(form, constrid) 
     end
-    return DualSolution(form, constrids, constrvals, bound)
+    return DualSolution(form, constrids, constrvals, bound, FEASIBLE_SOL)
```
guimarqu, line 108: @rrsadykov I'm not sure about this `SolutionStatus`. Does this solution require a status ?

src/Algorithm/colgenstabilization.jl
```diff
@@ -144,7 +144,7 @@ function update_alpha_automatically!(
 
     # first we calculate the in-sep direction
     constrids, constrvals = componentwisefunction(smooth_dual_sol, storage.stabcenter, -)
-    in_sep_direction = DualSolution(master, constrids, constrvals, 0.0)
+    in_sep_direction = DualSolution(master, constrids, constrvals, 0.0, FEASIBLE_SOL)
```
guimarqu, line 147: Same question here.

src/Algorithm/colgenstabilization.jl
```diff
@@ -158,7 +158,7 @@ function update_alpha_automatically!(
             push!(constrrhs, getcurrhs(master, constrid))
         end 
     end
-    subgradient = DualSolution(master, constrids, constrrhs, 0.0)
+    subgradient = DualSolution(master, constrids, constrrhs, 0.0, FEASIBLE_SOL)
```
guimarqu, line 161: Same here.

src/Algorithm/colgenstabilization.jl
rrsadykov: You can put the solutions status to something like "unknown_feasibility" for the dual solutions used for stabilization.

guimarqu:
Ready to merge, waiting for approvation.

rrsadykov:
Ok for me. However, it seems I do not have rights to approve this merge.

guimarqu:
> Ok for me. However, it seems I do not have rights to approve this merge.

But you are in the list of possible reviewers.

rrsadykov (approved):

rrsadykov:
Ok, I found.

PR #402, opened 2020-10-25
Remove MoiResult & mv optimization of Formulations in Algorithm

guimarqu:
- remove useless methods of FormulationManager
- remove method `optimize!` from `MathProg`, the formulation is now solved using `SolveIpForm` or `SolveLpForm` algorithms
- remove `MoiResult`

For a future PR : refactoring of `SolveIpForm` & `SolveLpForm`


src/MathProg/MOIinterface.jl
```diff
@@ -182,15 +177,50 @@ function _getcolunakind(record::MoiVarRecord)
...
+function get_primal_solutions(form::F, optimizer::MoiOptimizer) where {F <: Formulation}
+    inner = getinner(optimizer)
+    nb_primal_sols = MOI.get(inner, MOI.ResultCount())
+    solutions = PrimalSolution{F}[]
+    for res_idx in 1:nb_primal_sols
         if MOI.get(inner, MOI.PrimalStatus(res_idx)) != MOI.FEASIBLE_POINT
             continue
         end
-        solcost = MOI.get(inner, MOI.ObjectiveValue())
+        solcost = 0.0 # TODO : constant in the objective function (e.g. partial solution fixed) ?
```
rrsadykov, line 223: MOI does not know about fixed partial solutions. Partial solutions are treated only by the algorithms (by concatenation of MOI solution and partial solution when needed)

src/MathProg/formulation.jl
```diff
@@ -6,6 +6,7 @@ mutable struct Formulation{Duty <: AbstractFormDuty}  <: AbstractFormulation
     optimizer::AbstractOptimizer
     manager::FormulationManager
     obj_sense::Type{<:Coluna.AbstractSense}
+    robust_constr_generators::Vector{RobustConstraintsGenerator}
```
rrsadykov, line 9: What are robust constraint generators?

rrsadykov (commented):
What is the reason to move optimizaton of formulations to Algorithm submodule?

src/Algorithm/basic/solveipform.jl
```diff
@@ -94,15 +97,50 @@ end
...
+    end
+    return termination_status!(result, optimizer, form)
 end
-check_if_optimizer_supports_ip(optimizer::UserOptimizer) = true
 
 function optimize_ip_form!(algo::SolveIpForm, optimizer::MoiOptimizer, form::Formulation)
     MOI.set(optimizer.inner, MOI.TimeLimitSec(), algo.time_limit)
     MOI.set(optimizer.inner, MOI.Silent(), algo.silent)
-    # No way to enforce upper bound through MOI.
+    # No way to enforce upper bound or lower bound through MOI.
```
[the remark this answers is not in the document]
guimarqu: Yes I know. I think the argument to not add this attribute was "you can do it with a constraint". I'll look for the discussion. I'll try to address that in the future SolveIpForm/SolveLpForm refactoring.

src/MathProg/MOIinterface.jl
guimarqu: Yes but I think we should move partial solutions in MathProg as a fixed part of the model. I'm going to remove the comment and we'll see that later.

src/Algorithm/basic/solvelpform.jl
```diff
@@ -45,7 +45,17 @@ end
...
 function optimize_lp_form!(algo::SolveLpForm, optimizer::MoiOptimizer, form::Formulation)
     MOI.set(form.optimizer.inner, MOI.Silent(), algo.silent)
-    return optimize!(form)
+    result = OptimizationState(form)
+    sols_found =  optimize_with_moi!(optimizer, form, result)
+    if sols_found
+        for primal_sol in get_primal_solutions(form, optimizer)
+            add_lp_primal_sol!(result, primal_sol)
+        end
+        for dual_sol in get_dual_solutions(form, optimizer)
```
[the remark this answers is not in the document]
guimarqu: Ok, I'll fix this in SolveIpForm/SolveLpForm refactoring.

src/MathProg/formulation.jl
guimarqu: These are the robust cut callbacks, they were in `FormulationManager` before. I don't know why I moved them here... I'm going to revert this.

src/MathProg/MOIinterface.jl
rrsadykov: I think partial solutions should remain a part of Algorithm. It can be expensive to concatenate the solution each time. Now we concatenate only if a better integer solution is found. For fractional solutions we do not care if they are concatenated or not.

rrsadykov (approved):

guimarqu:
> What is the reason to move optimizaton of formulations to Algorithm submodule?

To delete `MoiResult` & because we already have `SolveIpForm` and `SolveLpForm`.

PR #415, opened 2020-11-18
Increase max nb of formulations + better error message if var/constr …

guimarqu:
…exists

Fix for https://github.com/atoptima/BlockDecomposition.jl/pull/31

This PR also fixes :  
- completion of branching file
- tolerance for integer feasibility of a solution

### DIFF
--- a/src/MathProg/solutions.jl
+++ b/src/MathProg/solutions.jl
@@ -19,7 +19,7 @@ end
 function Base.isinteger(sol::Solution)
     for (vc_id, val) in sol
         #if getperenkind(sol.model, vc_id) != Continuous
-            !isinteger(val) && return false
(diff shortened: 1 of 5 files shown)


src/MathProg/solutions.jl
```diff
@@ -19,7 +19,7 @@ end
 function Base.isinteger(sol::Solution)
     for (vc_id, val) in sol
         #if getperenkind(sol.model, vc_id) != Continuous
-            !isinteger(val) && return false
+            abs(round(val) - val) <= 1e-5 || return false
```
rrsadykov, line 22: This tolerance should be a parameter, no?

guimarqu: Yes, we'll do this later. I opened an issue.

rrsadykov (approved):

PR #434, opened 2021-01-16
Fixing the branching tree file

artalvpes:
See the attached file as an example of the generated branching tree.
[cvrp2.pdf](https://github.com/atoptima/Coluna.jl/files/5824580/cvrp2.pdf)

### DIFF
--- a/src/Algorithm/treesearch.jl
+++ b/src/Algorithm/treesearch.jl
@@ -196,7 +196,7 @@ function init_branching_tree_file(algo::TreeSearchAlgorithm)
         open(algo.branchingtreefile, "w") do file
             println(file, "## dot -Tpdf thisfile > thisfile.pdf \n")
             println(file, "digraph Branching_Tree {")
-            println(file, "\tedge[fontname = \"Courier\", fontsize = 10];")
+            print(file, "\tedge[fontname = \"Courier\", fontsize = 10];}")
         end
     end
     return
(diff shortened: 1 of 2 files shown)


src/Algorithm/treesearch.jl
```diff
@@ -206,13 +206,20 @@ function print_node_in_branching_tree_file(algo::TreeSearchAlgorithm, data::Tree
     if algo.branchingtreefile !== nothing
         pb = getvalue(get_ip_primal_bound(getoptstate(data)))
         db = getvalue(get_ip_dual_bound(getoptstate(node)))
+        write(algo.branchingtreefile, read(algo.branchingtreefile, String)[1:end-1])
```
guimarqu, line 209: Do you want to remove the last line ? Any better way to do this ?

artalvpes: No. Only the last character (a closing brace). I pushed a change to use chop() instead.

guimarqu: Is it possible to handle the closing brace in another way than copying (almost) the whole file ?

guimarqu:
Alternative ?

```julia
io = open("file", "r+")
seekend(io)
truncate(io, position(io)-1)  # assert that position(io) >= 1,   remove last character (last byte actually)
write!(io, "something")
close(io)
```

cf https://docs.julialang.org/en/v1/base/io-network/

guimarqu:
> Alternative ?
> 
> ```julia
> io = open("file", "r+")
> seekend(io)
> truncate(io, position(io)-1)  # assert that position(io) >= 1,   remove last character (last byte actually)
> write!(io, "something")
> close(io)
> ```
> 
> cf https://docs.julialang.org/en/v1/base/io-network/

Even better, you don't need to truncate, you just seek for position `end - 1` and you start writing there : 
```julia
seekend(io)
n = position(io)
seek(io, n-1)
write(io, "something")
```

guimarqu (approved):

PR #437, opened 2021-01-28
Basic & MIP unit tests for MOI wrapper

guimarqu:
Improve #436 

TODO list : 

- [x] #425
- [x] support of PrimalStatus
- [x] support of DualStatus
- [x] support of ConstraintPrimal
- [x] support of ResultCount
- [x] get constraint by name
- [x] get variable by name
- [x] support of ObjectiveSense
- [x] support of FeasbilitySense 
- [x] support of RawStatusString

It fixes bugs with binary variables in the subsolver

Issue #301, opened 2020-03-20
Is Coluna possible to call solvers (e.g., Cplex, Gurobi) directly?


YP-Ye:
This problem may be a little strange. I have met a problem that I want to solve by Cplex or Gurobi in iterations 1-5, and then solve it by decomposition methods in the following iterations. Can it be realized by Coluna? I have not found the option to use Cplex or Gurobi directly.

guimarqu:
Hi,

I’m not sure to understand what you want to do. 

One option is to create your model twice and apply the decomposition on the second one. You solve the first model using a MIP solver and the second one using Coluna.

A second option is to create a new algorithm that will solve the original formulation (it’s a field of `Reformulation`) using `SolveIpForm` and then call `ColGen`. There is an example in `tests/unit/algorithms/masteripheur.jl` where I create an algorithm. There is no doc at the moment and the algorithm submodule is a little bit messy.

guimarqu:
The answer to the main question is yes.

YP-Ye:
> Hi,
> 
> I’m not sure to understand what you want to do.
> 
> One option is to create your model twice and apply the decomposition on the second one. You solve the first model using a MIP solver and the second one using Coluna.
> 
> A second option is to create a new algorithm that will solve the original formulation (it’s a field of `Reformulation`) using `SolveIpForm` and then call `ColGen`. There is an example in `tests/unit/algorithms/masteripheur.jl` where I create an algorithm. There is no doc at the moment and the algorithm submodule is a little bit messy.

Thanks for your kindly reply. In my model, I add some constraints step by step, and in each step I want to use decomposition methods or Cplex directly. In the first step, I met a error:
`MethodError: no method matching store!(::Coluna.MathProg.Annotations, ::Nothing, ::Coluna.MathProg.Constraint)
`
I do not know whether it is a bug or it just cannot be solved by Coluna.

guimarqu:
At the moment, Coluna does not support cut generation (it’s the next main feature we will implement) nor block generation at the user level.

How do you add your constraints ? It’s a cut generation or do you want to generate new subproblems during the optimization ?

Do you want to modify the model and restart the optimization ?

YP-Ye:
> At the moment, Coluna does not support cut generation (it’s the next main feature we will implement) nor block generation at the user level.
> 
> How do you add your constraints ? It’s a cut generation or do you want to generate new subproblems during the optimization ?
> 
[...]

Issue #223, opened 2019-12-18
Relax master integrality in algorithm


guimarqu:
In file `optimize.jl` , initialisation of Coluna, we relax the integrality of the master. 
It should be done in algorithms.

guimarqu:
closed by #438

PR #438, opened 2021-01-29
no reformulation when user does not provide decomposition

### LINKED ISSUE #425: Do not reformulate a non-decomposed model
Related to #424 (same error msg)

guimarqu:
fix #425 #223 

improvement for #301 

needed for #437

### DIFF
--- a/test/issues_tests.jl
+++ b/test/issues_tests.jl
@@ -1,0 +1,22 @@
+
+function test_issues_fixed()
+    # Issue 425
+    @testset "#425" begin
+        coluna = JuMP.optimizer_with_attributes(
+            Coluna.Optimizer,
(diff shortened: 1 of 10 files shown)


test/issues_tests.jl
```diff
@@ -0,0 +1,22 @@
+
+function test_issues_fixed()
```
vitornesello, line 2: Impossible to understand what the tests does without going to github and searching for the issue

vitornesello: I think it should at least have a readable name

guimarqu: Ok and I'm going to fix the doc

vitornesello (approved):

PR #444, opened 2021-02-06
Fixed a bug when restoring master cuts in strong branching

artalvpes:
Note: I have added several (disabled) debugging tools used to find the bug.

### DIFF
--- a/src/Algorithm/basic/cutcallback.jl
+++ b/src/Algorithm/basic/cutcallback.jl
@@ -27,7 +27,8 @@ end
 
 # CutCallbacks does not have child algorithms, therefore get_child_algorithms() is not defined
 
-function get_storages_usage(algo::CutCallbacks, form::Formulation{MathProg.AbstractMasterDuty}) 
+function get_storages_usage(algo::CutCallbacks, form::Formulation{Duty}
+    ) where {Duty<:MathProg.AbstractFormDuty} 
     return [(form, MasterCutsStoragePair, READ_AND_WRITE)]
 end
 
--- a/src/Algorithm/basic/solvelpform.jl
+++ b/src/Algorithm/basic/solvelpform.jl
@@ -47,12 +47,7 @@ function optimize_lp_form!(
     algo::SolveLpForm, optimizer::MoiOptimizer, form::Formulation, result::OptimizationState
 )
     MOI.set(form.optimizer.inner, MOI.Silent(), algo.silent)
-    # MOI.set(form.optimizer.inner, MOI.RawParameter("Method"), 2)
     optimize_with_moi!(optimizer, form, result)
-    #Coluna.inc_lp_file_count()
-    #lpfile = "coluna_master$(Coluna.lp_file_count).lp"
-    #println("Writing LP file $lpfile")
-    #Gurobi.GRBwrite(form.optimizer.inner, lpfile)
     return
 end
 
--- a/src/Algorithm/basic/solvelpform.jl
+++ b/src/Algorithm/basic/solvelpform.jl
@@ -47,7 +47,12 @@ function optimize_lp_form!(
     algo::SolveLpForm, optimizer::MoiOptimizer, form::Formulation, result::OptimizationState
 )
     MOI.set(form.optimizer.inner, MOI.Silent(), algo.silent)
+    # MOI.set(form.optimizer.inner, MOI.RawParameter("Method"), 2)
     optimize_with_moi!(optimizer, form, result)
+    #Coluna.inc_lp_file_count()
+    #lpfile = "coluna_master$(Coluna.lp_file_count).lp"
+    #println("Writing LP file $lpfile")
+    #Gurobi.GRBwrite(form.optimizer.inner, lpfile)
     return
 end
 
--- a/src/Algorithm/branching/branchingalgo.jl
+++ b/src/Algorithm/branching/branchingalgo.jl
@@ -178,10 +178,6 @@ function perform_strong_branching_with_phases!(
 
         sort!(groups, rev = true, by = x -> (x.isconquered, x.score))
 
-        if groups[1].isconquered
-            nb_candidates_for_next_phase == 1 
-        end
-
         # before deleting branching groups which are not kept for the next phase
         # we need to remove storage states kept in these nodes
         for group_index = nb_candidates_for_next_phase + 1 : length(groups) 
--- a/src/Algorithm/branching/branchingalgo.jl
+++ b/src/Algorithm/branching/branchingalgo.jl
@@ -178,6 +178,10 @@ function perform_strong_branching_with_phases!(
 
         sort!(groups, rev = true, by = x -> (x.isconquered, x.score))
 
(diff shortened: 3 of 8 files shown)


src/Algorithm/branching/branchingalgo.jl
```diff
@@ -178,10 +178,6 @@ function perform_strong_branching_with_phases!(
 
         sort!(groups, rev = true, by = x -> (x.isconquered, x.score))
 
-        if groups[1].isconquered
-            nb_candidates_for_next_phase == 1 
-        end
```
rrsadykov, line 183: Artur, why did you remove this?

src/Algorithm/basic/cutcallback.jl
```diff
@@ -27,7 +27,8 @@ end
 
 # CutCallbacks does not have child algorithms, therefore get_child_algorithms() is not defined
 
-function get_storages_usage(algo::CutCallbacks, form::Formulation{MathProg.AbstractMasterDuty}) 
+function get_storages_usage(algo::CutCallbacks, form::Formulation{Duty}
+    ) where {Duty<:MathProg.AbstractFormDuty}
```
rrsadykov, line 31: Nice! Sorry for this bug.

src/Coluna.jl
```diff
@@ -21,6 +21,40 @@ const TOL = 1e-8 # if - ϵ_tol < val < ϵ_tol, we consider val = 0
...
+function inc_debug_count()
+    global debug_count += 1
+    return
+end
+function debug_stack()
+    for l in stacktrace()
+        println(l)
+    end
+end
+macro debug()
```
rrsadykov, line 35: Is it possible to add a debug parameter and a test to cover these debug functions.

rrsadykov (commented):
Artur, can you explain how your debugging tools work?

src/Coluna.jl
guimarqu: The debug tool should not stay in the code otherwise we need to test it. Can we move it to the documentation ?

src/Algorithm/basic/solvelpform.jl
```diff
@@ -47,7 +47,12 @@ function optimize_lp_form!(
     algo::SolveLpForm, optimizer::MoiOptimizer, form::Formulation, result::OptimizationState
 )
     MOI.set(form.optimizer.inner, MOI.Silent(), algo.silent)
+    # MOI.set(form.optimizer.inner, MOI.RawParameter("Method"), 2)
```
guimarqu, line 50: Can we remove these comments ?

src/MOIcallbacks.jl
```diff
@@ -100,7 +100,7 @@ function MOI.submit(
     end
 
     constr = setconstr!(
-        form, "", MasterMixedConstr;
+        form, "", MasterBendCutConstr; #TODO: Create a specific type for user cuts
```
guimarqu, line 103: Why did you change the duty ?

src/Algorithm/branching/branchingalgo.jl
artalvpes: Hi Ruslan. I think it had no effect. Did you intend to use = instead of == ?

src/Coluna.jl
artalvpes: No problem. I just did not want to loose it.

src/Algorithm/basic/solvelpform.jl
artalvpes: No problem.

src/MOIcallbacks.jl
artalvpes: This is important. I had to fix two things to make it work. This is one of them. The user cuts were not being recognized by the constructor of MasterCutsState because of the condition "getduty(id) <= AbstractMasterCutConstr" and I did not want to change the complex hierarchy defined in "duties.jl". So defining it as a MasterBendCutConstr was a solution with minimum changes to avoid generating new bugs.

src/Algorithm/branching/branchingalgo.jl
artalvpes: Anyway, we decided to let the nodes survive a little bit more before being deleted to let them appear in the branch tree file. We don't think there will be a significant impact in the performance.

artalvpes:
I think it is worth letting a general comment about the code structure here. I think it would help a lot for people that starts looking at the code if there was a more clearly marked standard structure for the storages that were followed by all algorithms. Perhaps we could put each algorithm and each storage type in a separate file and force all storage functions to be redefined (like get_storages_usage and get_child_algorithms) even if they do nothing (by generating an error message in the function that receives the abstract type). Calls to restore_states!, store_states! and remove_states! could perhaps be more emphasised and/or we could try to keep them in the first call level (function run!) of their algorithms.

src/Algorithm/branching/branchingalgo.jl
rrsadykov: Yes, it should be =.  This is needed to avoid unnecessary evaluation of branching candidates, when one of candidates is already conquered.

rrsadykov:
I agree that we need a document describing the architecture of Coluna (environment, models, formulations, algorithms, storages, and other concepts we use). Guillaume should have started such a document. I will review and complete it when Guillaume is done with his part.  A link to this document should be in the README.md file.

src/Algorithm/branching/branchingalgo.jl
artalvpes: I suggest marking it as conquered and testing it in the beginning of the evaluation to avoid evaluating it again but letting it persist until being written to the branch tree file (equivalent to BaPTree.dot). Otherwise some nodes will appear with only one child.

rrsadykov: I did not completely get your point. Branching candidate (group) here is not a child node, but a complete set of child nodes (left branch and right branch).  The group is already marked as conquered and will not be evaluated again.  This is to avoid passing other candidates (or groups) to the next strong branching phase. Thus we avoid evaluating them. This evaluation is not needed as there exists a conquered branching candidate, i.e. all children nodes are conquered. We pass only this branching candidate, and all child nodes of this candidate will appear in the branch tree file.

artalvpes: Ah! ok. Got it. So, we should put the code back with = replacing ==.

src/Algorithm/storage.jl
```diff
@@ -90,6 +90,13 @@ const StorageTypePair = Pair{DataType, DataType}
 #         be restored for writing (all other storages are restored anyway but just for reading)
 const StoragesUsageDict = Dict{Tuple{AbstractModel, StorageTypePair}, StorageAccessMode}
 
+function Base.show(io::IO, stoUsaDict::StoragesUsageDict)
```
guimarqu, line 93: Variable name must be lower case.

guimarqu (changes requested):
One style issue to address and then the PR is ready to be merged.

guimarqu (approved):

PR #447, opened 2021-02-22
test empty model solution and improve error msg

laradicp:


test/issues_tests.jl
```diff
@@ -73,6 +73,33 @@ function decomposition_with_constant_in_objective()
     @test objective_value(model) ≈ 307.5 + 2
 end
 
+# Issue #424
+# Expected behaviors when trying to solve an empty model
```
guimarqu, line 77: I would describe the behaviors in comments  : 
-If you try to solve an empty model with Coluna using a `SolveIpForm` or `SolveLpForm` as top solver, the objective value will be 0. 
-If you try to solve an empty model using `TreeSearchAlgorithm`, then Coluna will throw an error because since there is no decomposition, there is no reformulation and `TreeSearchAlgorithm` must be run on a reformulation.

test/issues_tests.jl
```diff
@@ -73,6 +73,33 @@ function decomposition_with_constant_in_objective()
...
+function solve_empty_model()
+    coluna = JuMP.optimizer_with_attributes(
+        Coluna.Optimizer,
+        "params" => CL.Params(solver = ClA.SolveIpForm()),
+        "default_optimizer" => GLPK.Optimizer
+    )
+    model = BlockModel(coluna, direct_model = true)
+    optimize!(model)
+    @test JuMP.objective_value(model) == Inf
```
guimarqu, line 86: The objective value must be 0.

test/issues_tests.jl
```diff
@@ -73,6 +73,33 @@ function decomposition_with_constant_in_objective()
...
+            solver = Coluna.Algorithm.TreeSearchAlgorithm()
+        ),
+        "default_optimizer" => GLPK.Optimizer
+    )
+    model = BlockModel(coluna)
+    try
+        optimize!(model)
+    catch e
+        @test repr(e) == "ErrorException(\"Cannot apply run! for arguments Coluna.Algorithm.TreeSearchAlgorithm, Coluna.Algorithm.ModelData, Coluna.Algorithm.OptimizationInput{Coluna.MathProg.Formulation{Coluna.MathProg.Original},Coluna.MathProg.MinSense}.\")"
+    end
```
guimarqu, line 100: You can replace this block by `@test_throws ExceptionError optimize!(model)`.

guimarqu (commented):
I don't understand why coverage decreases...

guimarqu (approved):

PR #457, opened 2021-03-01
Add better error msg if max number of formulations reached

laradicp:


src/MathProg/formulation.jl
```diff
@@ -26,6 +26,9 @@ function Formulation{D}(
     parent_formulation = nothing,
     obj_sense::Type{<:Coluna.AbstractSense} = MinSense
 ) where {D<:AbstractFormDuty}
+    if form_counter.value ≥ MAX_NB_FORMULATIONS
```
guimarqu, line 29: You should use `>=` rather than `≥`.

guimarqu (approved):

PR #462, opened 2021-03-05
Optimize twice

laradicp:


guimarqu:
Let's focus on reformulation with `direct_model = true`, it looks like there is a bug because we must reinitialize parts of the `Annotations` data structure. This structure is used to store to which subproblem the constraints and variables belong. It is initialized when we call `Coluna.Optimizer`. Then, it is populated when BlockDecomposition assigns variables and constraints to subproblems, for instance : 
https://github.com/atoptima/Coluna.jl/blob/4d327613aa0a146fd010374c7fa07cc019b76163/src/MOIwrapper.jl#L378-L385

and it is populated a last time when Coluna does the reformulation (calls to `store!` in `buildformulation`).

To fix the bug, you must keep what is stored in MOI wrapper (because the model passes through MOI only once when `direct_model = true`) and erase what is stored in `reformulate!`. I think only `ann_per_form` needs to be reinitialized.

guimarqu:
I think for the first (with no reformulation and `direct_model=true`), when you write the model, the original formulation is stored in the formulation and the buffer of the subsolver. When you call `optimize!`, the content of the buffer is pushed in the subsolver and then erased. When you call `optimize!` a second time, the buffer is empty and thus the subsolver solves an empty formulation.

guimarqu (commented):
Since `optimize!(model)` returns `nothing`, you don't test if it's really working. You should retrieve the objective value & the solution of the two runs and compare them.

For the reformulation, you can use the generalized assignment (using the call to ColunaDemos such as in full instance tests).

guimarqu (approved):

Issue #330, opened 2020-05-02
Names for Data/Storage/StorageState


guimarqu:
I don't understand how you can restore the state of an empty storage ? The state should be empty ? Maybe we should rename `AbstractStorageState` to `AbstractDataState` because it's the state of the `AbstractData` ?

_Originally posted by @guimarqu in https://github.com/atoptima/Coluna.jl/pull/323_

waiting to get some ideas.

----

EDIT (2021/03/01), 2nd EDIT (2021/03/18): 

Todo: 

- [x] `StorageState` becomes `Record`
- [x] `Storage` becomes `(Storage)Unit`
- [x] change order of arguments in `restore_from_record!`
- [x] `StorageContainer` becomes `StorageUnitWrapper`
- [x]  move storage into `ColunaBase`
- [x]  move `StorageDict` into  `Model`, then delete `AbstractData` (new method in model interface)
- [x] `StorageDict` becomes `Storage`
- [ ] unit tests on storage "public" methods

Decisions to make : 
- [ ] method to instanciate records ? https://github.com/atoptima/Coluna.jl/pull/470#discussion_r596941248
- [ ] method to instanciate units ?
- [ ] remove `NOT_USED`   (because `NOT_USED` and `READ_ONLY` are the same) (@rrsadykov)

Improve dev Exp : 
- [ ] `RecordsVector` -> `Vector{Record}`
- [x] `StorageDict`  ? -> `Storage`
- [x] `UnitTypePair = Pair{DataType, DataType}` ? -> removed because now bijection between UnitType & RecordType
- [x] `UnitsUsageDict = Dict{Tuple{AbstractModel, UnitTypePair}, UnitAccessMode}` ? -> `UnitsUsage`

Conversations to resolve : 
- [ ] https://github.com/atoptima/Coluna.jl/pull/471#discussion_r597499436
- [ ] https://github.com/atoptima/Coluna.jl/pull/472#discussion_r597623524
- [ ] https://github.com/atoptima/Coluna.jl/pull/518#discussion_r620311675

Variables to rename : 
- [ ]  storage / storage unit wrapper in storages.jl
- [ ]  storagedict / storage 

Needs documentation (or remove)
- [ ] `initialize_storage_units`
- [ ] `collect_units_to_restore!` (why a UnitsUsageDict as first arg ?)
- [ ] `store_records(reform)`
- [ ] `store_records(reform, record_vector)`
- [ ] `getstorageunit(model, storage_unit_type)`
[...]

PR #470, opened 2021-03-18
rename StorageState -> RecordState

guimarqu:
task for #330

### DIFF
--- a/src/Algorithm/storage.jl
+++ b/src/Algorithm/storage.jl
@@ -9,15 +9,15 @@
     Models are storages themselves. Each storage is associated with a
     model. Thus a storage adds computed data to a model.  
 
-    Storage states are useful to store states of storages at some point 
+    Record states are useful to store states of storages at some point 
     of the calculation flow so that we can later return to this point and 
     restore the storages. For example, the calculation flow may return to
     some saved node in the search tree.
 
     Some storages can have different parts which are stored in different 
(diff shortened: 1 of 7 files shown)


src/Algorithm/storage.jl
```diff
@@ -45,44 +45,44 @@
...
-    takes a model and a storage as parameters. This constructor
-    is called during storing a storage. 
+    AbstractRecordState
+
+A record state is the particular condition that a record is in at a specific time
+of the execution of Coluna.
     
+For each record state, a constructor should be defined which
+takes a model and a storage as parameters. This constructor
+is called during storing a storage.
```
guimarqu, line 55: @rrsadykov Do you think we can define a method `newstate(model, record)` in the storage interface that will return a record state that corresponds to the model and the record ?

rrsadykov: In principle, we can do that, I have no preference for a function or a constructor. However, sometimes from the same record, we can produce states of different type, i.e. from formulation, we can produce MasterColumnsRecordState and MasterCutsColumnState. Such a function should be able to specify that.

rrsadykov (approved):

src/Algorithm/storage.jl
guimarqu: Ok, I'll put a permalink to this conversation in #330. Let's decide later.

Issue #397, opened 2020-10-14
update_all_ip_primal_solutions! reverse loop


guimarqu:
https://github.com/atoptima/Coluna.jl/blob/50b2f4ed95dc7cfa95f4cea5cc9bddedfeefd79e/src/Algorithm/utilities/optimizationstate.jl#L223-L231

@rrsadykov what does it change if you loop in reverse or normal mode ? I don't understand the comment.

guimarqu:
Because you want to store all of them and if you start by the beginning you'll only store the best one ?

rrsadykov:
Yes, this is the idea: we use `update_ip_primal_sol!()` to store only improving solutions. Otherwise the same solution may be stored several times (`orig_state` is usually initialized by `dest_state` in the code), or we would need to compare all solutions one-by-one which takes time.
Because we use `update_ip_primal_sol!()`, we need to call it for new solutions in the reverse order, otherwise only the best improving solution would be stored.

guimarqu:
Ok nice

guimarqu:
I leave the issue open, I would like to try things later to see if we can avoid having too many extra methods to handle solutions

guimarqu:
This method does not exist anymore.

PR #474, opened 2021-03-20
Retrieve multiple solutions from `SolveIpForm`

guimarqu:
@artalvpes  This is the fix to generate multiple column each time the pricing callback is run.

Changes : 
- A new parameter in `SolveIpForm` to define the maximum number of solution returned
- Fix the way we retrieve solutions in SolveIpForm
- Change for `(update|add)_(ip|lp)_(primal|dual)_sol` : now we create a new entry in the list of solutions, sort the list, and then the less good solutions are deleted to respect the size of the list (method `bestbound!`). Before this PR, the new solution was replacing a solution of the list even if the replaced solution was better than the new one. (@rrsadykov I think it may allow us to remove `update_all_ip_primal_solutions!` cf #397).
- `set_(lp|ip)_(primal|dual)_sol` -> `unshift_(lp|ip)_(primal|dual)_sol` because I think `set` is misleading when the list has a size > 1. The new method pushes the new solution at the beginning of the list. The new solution is therefore considered as the best solution.

(@ItamarRocha, lot of things related to solutions in this PR)

### DIFF
--- a/docs/src/dev/todo.md
+++ b/docs/src/dev/todo.md
@@ -49,7 +49,7 @@ Coluna.MathProg.update_lp_primal_bound!
 Coluna.MathProg.Variable
 Coluna.MathProg.MoiOptimizer
 Coluna.ColunaBase.Bound
-Coluna.Algorithm.set_ip_primal_sol!
+Coluna.Algorithm.unshift_ip_primal_sol!
 Coluna.MathProg.get_ip_dual_bound
 Coluna.Algorithm.ReformData
 Coluna.Algorithm.ModelData
--- a/docs/src/dev/todo.md
+++ b/docs/src/dev/todo.md
@@ -49,7 +49,7 @@ Coluna.MathProg.update_lp_primal_bound!
 Coluna.MathProg.Variable
 Coluna.MathProg.MoiOptimizer
 Coluna.ColunaBase.Bound
-Coluna.Algorithm.unshift_ip_primal_sol!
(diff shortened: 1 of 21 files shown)


rrsadykov:
I agree that the `set_sol` is misleading, but the word `unshift` does not exist and thus even more misleading, i think

guimarqu:
> I agree that the `set_sol` is misleading, but the word `unshift` does not exist and thus even more misleading, i think

I think I should replace by `pushfirst` because it's the name of the method in julia. A lot of languages has a method `unshift` to push something at the beginning of an array.

rrsadykov:
We may leave `set_sol`, but when this function is used, all previous solutions are removed. `set_sol` would be used for the case when the can be only one solution. For example, in column generation, there is only one dual and one primal solution.

guimarqu:
> We may leave `set_sol`, but when this function is used, all previous solutions are removed. `set_sol` would be used for the case when there can be only one solution. For example, in column generation, there is only one dual and one primal solution.

That was my 2nd option, so it's good for me.

rrsadykov:
I think there are two distinct use cases: one we have one solution (`set_sol` is ok for this case). Other case is when we have a list of solutions sorted by bound.  In the latter case, just `add_sol` would be enough i think. Are there any other use cases?

guimarqu:
I also see these two cases. We provide `add_sol` and `update_sol`. The second one adds the solution to the list only if the solution improves the bound.

rrsadykov (approved):

PR #475, opened 2021-03-22
Remove code for spMv operation in colgen

### LINKED ISSUE #311: Move spMv operation in DynamicSparseArrays
https://github.com/atoptima/Coluna.jl/pull/309#issuecomment-614140127

guimarqu:
- close #311 
- waiting for release of DynamicSparseArrays v0.4.0 (already implemented https://github.com/atoptima/DynamicSparseArrays.jl/pull/18) ✅ 
- close #501 

--

**Performance decrease of reduced costs computation** : for the pricing subproblems, we only need to compute the reduced costs of the subproblem representative variables. Here, we iterate over all the non-zero values of the coefficient matrix so we can also compute the reduced cost of each column (waste of time).

Solutions : 
1. Separate subproblem representative variables from other variables **in col gen algo only**. ~~We maintain them in another matrix. It will need investigation to see what we should do for Benders.~~ (cf Ruslan's comment)
2. Status quo until it becomes critical
3. Other ? It's open to discussion (@artalvpes @rrsadykov  @vitornesello @FD-V)

PS : not urgent.

### DIFF
--- a/src/Algorithm/colgen.jl
+++ b/src/Algorithm/colgen.jl
@@ -66,23 +66,53 @@ function get_units_usage(algo::ColumnGeneration, reform::Reformulation)
     return units_usage
 end
 
-struct ReducedCostsVector
+struct ReducedCostsCalculationHelper
     length::Int
-    varids::Vector{VarId}
+    dwspvarids::Vector{VarId}
     perencosts::Vector{Float64}
-    form::Vector{Formulation}
+    dwspforms::Vector{Formulation}
+    dwsprep_coefmatrix::MathProg.ConstrVarMatrix
 end
 
-function ReducedCostsVector(varids::Vector{VarId}, form::Vector{Formulation})
-    len = length(varids)
-    perencosts = zeros(Float64, len)
-    p = sortperm(varids)
-    permute!(varids, p)
-    permute!(form, p)
-    for i in 1:len
-        perencosts[i] = getcurcost(getmaster(form[i]), varids[i])
+# Pre compute information to speed-up calculation of reduced costs of original variables
+# in the master of a given reformulation.
+function ReducedCostsCalculationHelper(reform::Reformulation)
+    dwspvarids = VarId[]
+    dwspforms = Formulation[]
+
+    for (spid, spform) in get_dw_pricing_sps(reform)
+        for (varid, var) in getvars(spform)
+            if iscuractive(spform, varid) && getduty(varid) <= AbstractDwSpVar
+                push!(dwspvarids, varid)
+                push!(dwspforms, spform)
+            end
+        end
     end
-    return ReducedCostsVector(len, varids, perencosts, form)
+
+    len = length(dwspvarids)
+    perencosts = zeros(Float64, len)
+    p = sortperm(dwspvarids)
+    permute!(dwspvarids, p)
+    permute!(dwspforms, p)
+
+    master = getmaster(reform)
+    for i in 1:len
+        perencosts[i] = getcurcost(master, dwspvarids[i])
+    end
+
+    master_coefmatrix = getcoefmatrix(master)
+    dwsprep_coefmatrix = dynamicsparse(ConstrId, VarId, Float64)
+    for (constrid, constr) in getconstrs(master)
+        for (varid, coeff) in @view master_coefmatrix[constrid, :]
+            if getduty(varid) <= AbstractMasterRepDwSpVar
+                dwsprep_coefmatrix[constrid, varid] = coeff
+            end
+        end
+    end
+    closefillmode!(dwsprep_coefmatrix)
+    return ReducedCostsCalculationHelper(
+        len, dwspvarids, perencosts, dwspforms, dwsprep_coefmatrix
+    )
 end
 
 function run!(algo::ColumnGeneration, env::Env, data::ReformData, input::OptimizationInput)::OptimizationOutput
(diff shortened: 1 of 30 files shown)


rrsadykov:
Actually, `coeffmatrix` for subproblem representative variables is constant during the whole column generation. So, we should retrieve it in the beginning of the column generation, filter it to leave only this subproblem representative variables, and save it locally. Then this matrix is used to calculate the reduced costs on every iteration (thus, reduced costs of columns and representative variables for other subproblems are not calculated). The matrix should be recalculated only when core cuts are added.

rrsadykov:
As we talked before, it would be good to retrieve reduced costs of columns directly from the LP solver. It should be faster than calculate them from the coefficient matrix as we do now during master clean up.

guimarqu:
Ok your solution is a much better version of 1. It sounds good to me.

For the master clean up, we can't retrieve reduced costs through MOI (https://github.com/jump-dev/MathOptInterface.jl/issues/776). Hopefully, the sparse matrix sparse vector multiplication should be quite fast and I think that the deletion of the columns will be the bottleneck.

rrsadykov:
> For the master clean up, we can't retrieve reduced costs through MOI ([jump-dev/MathOptInterface.jl#776](https://github.com/jump-dev/MathOptInterface.jl/issues/776)). Hopefully, the sparse matrix sparse vector multiplication should be quite fast and I think that the deletion of the columns will be the bottleneck.

Ok, if it is not the bottleneck, the reduced cost of a column can be calculated from the dual of its >= 0 lower bound constraint, as they write there.

src/Algorithm/colgen.jl
```diff
@@ -66,23 +66,53 @@ function get_units_usage(algo::ColumnGeneration, reform::Reformulation)
...
+            end
+        end
+    end
+
+    len = length(dwspvarids)
     perencosts = zeros(Float64, len)
-    p = sortperm(varids)
-    permute!(varids, p)
-    permute!(form, p)
+    p = sortperm(dwspvarids)
```
rrsadykov, line 94: Why we sort subproblem variables?

src/Algorithm/colgen.jl
```diff
@@ -324,64 +354,23 @@ function solve_sp_to_gencol!(
     return
 end
 
-function updatereducedcosts!(reform::Reformulation, redcostsvec::ReducedCostsVector, dualsol::DualSolution)
-    redcosts = deepcopy(redcostsvec.perencosts)
+function updatereducedcosts!(
+    reform::Reformulation, redcostshelper::ReducedCostsCalculationHelper, dualsol::DualSolution
+)
+    redcosts = deepcopy(redcostshelper.perencosts)
```
rrsadykov, line 360: Why it is `deepcopy`, not just `copy`, what is the difference?

src/Algorithm/colgen.jl
```diff
@@ -324,64 +354,23 @@ function solve_sp_to_gencol!(
...
-
-    constr_key_pos::Int = 1
-    next_constr_key_pos::Int = 2
-
-    row_start = 0
-    row_end = 0
-    row_pos = 0
 
-    terms = Dict{VarId, Float64}(id => 0.0 for id in redcostsvec.varids)
+    result = transpose(redcostshelper.dwsprep_coefmatrix) * getsol(dualsol)
```
rrsadykov, line 363: May be it is better to transpose during pre-computation, and not on every iteration?

src/Algorithm/colgen.jl
```diff
@@ -66,23 +66,53 @@ function get_units_usage(algo::ColumnGeneration, reform::Reformulation)
...
+    dwsprep_coefmatrix = dynamicsparse(ConstrId, VarId, Float64)
+    for (constrid, constr) in getconstrs(master)
+        for (varid, coeff) in @view master_coefmatrix[constrid, :]
+            if getduty(varid) <= AbstractMasterRepDwSpVar
+                dwsprep_coefmatrix[constrid, varid] = coeff
+            end
+        end
     end
-    return ReducedCostsVector(len, varids, perencosts, form)
+    closefillmode!(dwsprep_coefmatrix)
```
rrsadykov, line 112: What `closefillmode!` serves for?

rrsadykov (changes requested):
I think you forgot to recalculate `dwsprep_coefmatrix` after adding core cuts

rrsadykov:
Ok, I saw that you quit `cg_main_loop!` after generating essential cuts, so `dwsprep_coefmatrix` will be updated.

However, I do not understand why you need to interrupt column generation loop after generating cuts?

src/Algorithm/colgen.jl
guimarqu: Indeed, it's not useful anymore because the ouput of spMv is a `Dict`.

src/Algorithm/colgen.jl
guimarqu: By default, when you fill a matrix of DynamicSparseArrays, all non-zero entries are stored in a buffer (writing the buffer is a constant time operation). Once you finish to fill the matrix, method `closefillemode!` flushes the buffer (complexity is the number of non-zero entries) and after that, a writing operation costs log^2(n) where n is number of non zero coefficients (if I remember well).

src/Algorithm/colgen.jl
guimarqu: AFAIK : 
- `copy` is a shallow copy (copy by value for immutable objects & copy by reference for mutable objects).
- `deepcopy` is a copy by value.

src/Algorithm/colgen.jl
guimarqu: It costs nothing

guimarqu:
> Ok, I saw that you quit `cg_main_loop!` after generating essential cuts, so `dwsprep_coefmatrix` will be updated.
> 
> However, I do not understand why you need to interrupt column generation loop after generating cuts?

Indeed, it does not make sense to restart colgen when it is in phase 1 or 3. I'll investigate in a new PR today.

rrsadykov (approved):

Issue #268, opened 2020-02-27
Create the original formulation and then the Problem


guimarqu:
Steps :

- create the original formulation
- get the annotations
- create the Problem when we decompose

When `optimize` is called a second time : 
- update the original formulation
- update the annotations
- create the Problem for the new decomposition

We decompose from scratch each time the original formulation is updated (i.e. MOI.optimize! called).

guimarqu:
@laradicp what do you think of this issue ? You worked on the "optimize twice" and it's working now.

laradicp:
> @laradicp what do you think of this issue ? You worked on the "optimize twice" and it's working now.

There's still this bug related to "optimize twice": https://github.com/atoptima/Coluna.jl/blob/ec1fa6b6893721d9ae26f1db8a964de2cda27fcc/test/MathOptInterface/MOI_wrapper.jl#L248
I don't know why it doesn't work, but maybe this could be a fix.

laradicp:
I don't think so, the sense is actually set correctly (I checked the Coluna sense after MOI.set and it was correct).

PR #484, opened 2021-03-26
Update Formulation docstring

laradicp:
task for #486

### DIFF
--- a/src/MOIwrapper.jl
+++ b/src/MOIwrapper.jl
@@ -17,7 +17,6 @@ const SupportedConstrSets = Union{
 mutable struct Optimizer <: MOI.AbstractOptimizer
     inner::Problem
     objective_type::ObjectiveType
-    params::Params
     annotations::Annotations
     #varmap::Dict{MOI.VariableIndex,VarId} # For the user to get VariablePrimal
     vars::CleverDicts.CleverDict{MOI.VariableIndex, Variable}
@@ -32,14 +31,13 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     default_optimizer_builder::Union{Nothing, Function}
 
     feasibility_sense::Bool # Coluna supports only Max or Min.
-
-    kpis::Union{Nothing, Kpis}
+    
+    env::Env
 
 
     function Optimizer()
         model = new()
         model.inner = Problem()
-        model.params = Params()
         model.annotations = Annotations()
         model.vars = CleverDicts.CleverDict{MOI.VariableIndex, Variable}()
         model.varids = CleverDicts.CleverDict{MOI.VariableIndex, VarId}() # TODO : check if necessary to have two dicts for variables
@@ -51,7 +49,7 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
         model.names_to_constrs = Dict{String, MOI.ConstraintIndex}()
         model.default_optimizer_builder = nothing
         model.feasibility_sense = false
-        model.kpis = nothing
+        model.env = Env(Params())
         return model
     end
 end
@@ -69,7 +67,7 @@ MOI.supports(::Optimizer, ::MOI.ConstraintDualStart) = false
 # Parameters
 function MOI.set(model::Optimizer, param::MOI.RawParameter, val)
     if param.name == "params"
-        model.params = val
+        model.env.params = val
     elseif param.name == "default_optimizer"
         optimizer_builder = () -> MoiOptimizer(MOI._instantiate_and_check(val))
         model.default_optimizer_builder = optimizer_builder
@@ -98,8 +96,8 @@ end
 MOI.get(optimizer::Coluna.Optimizer, ::MOI.SolverName) = "Coluna"
 
 function MOI.optimize!(optimizer::Optimizer)
-    optimizer.result, optimizer.kpis = optimize!(
-        optimizer.inner, optimizer.annotations, optimizer.params
+    optimizer.result = optimize!(
+        optimizer.env, optimizer.inner, optimizer.annotations
     )
     return
 end
@@ -674,8 +672,8 @@ function MOI.get(optimizer::Optimizer, ::MOI.ConstraintPrimal, index::MOI.Constr
     return constraint_primal(best_primal_sol, getid(constrid))
 end
 
-MOI.get(optimizer::Optimizer, ::MOI.NodeCount) = optimizer.kpis.node_count
-MOI.get(optimizer::Optimizer, ::MOI.SolveTime) = optimizer.kpis.elapsed_optimization_time
+MOI.get(optimizer::Optimizer, ::MOI.NodeCount) = optimizer.env.kpis.node_count
+MOI.get(optimizer::Optimizer, ::MOI.SolveTime) = optimizer.env.kpis.elapsed_optimization_time
 
 # function MOI.get(optimizer::Optimizer, ::MOI.ConstraintDual, index::MOI.ConstraintIndex)
 #     return 0.0
--- a/src/MOIwrapper.jl
+++ b/src/MOIwrapper.jl
@@ -15,6 +15,7 @@ const SupportedConstrSets = Union{
 
 @enum(ObjectiveType, SINGLE_VARIABLE, SCALAR_AFFINE)
 mutable struct Optimizer <: MOI.AbstractOptimizer
+    env::Env
     inner::Problem
     objective_type::ObjectiveType
     annotations::Annotations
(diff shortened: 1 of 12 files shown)


guimarqu:
Do you think we can improve this method ?

Something like : 

```julia
create_formulation!(problem::Problem, duty::Type{<:AbstractFormDuty}; 
     obj_sense = MinSense, parent_formulation = nothing)
```

laradicp:
> Do you think we can improve this method ?
> 
> Something like :
> 
> ```julia
> create_formulation!(problem::Problem, duty::Type{<:AbstractFormDuty}; 
>      obj_sense = MinSense, parent_formulation = nothing)
> ```
Yes, that makes sense to me. Should the method be kept in "formulation.jl" even if with the `::Problem` parameter?

guimarqu:
> > Do you think we can improve this method ?
> > Something like :
> > ```julia
> > create_formulation!(problem::Problem, duty::Type{<:AbstractFormDuty}; 
> >      obj_sense = MinSense, parent_formulation = nothing)
> > ```
> 
> Yes, that makes sense to me. Should the method be kept in "formulation.jl" even with the `::Problem` parameter?

Good question because otherwise we'll have cross-dep issue . Besides I would like to move problem outside MathProg (cf #439) so it's definitely not a good idea.

It's also does not look good to have Reformulation as first parameter.

Do you think we can move the formulation counter into the environment and then have : 

```julia
create_formulation!(env::Env, duty::Type{<:AbstractFormDuty}; 
     obj_sense = MinSense, parent_formulation = nothing)
```
?

It fits with the def of Env 

> Env is a place to store global information during the whole execution of Coluna. The environment contains the global parameter ...

laradicp:
> > > Do you think we can improve this method ?
> > > Something like :
> > > ```julia
> > > create_formulation!(problem::Problem, duty::Type{<:AbstractFormDuty}; 
> > >      obj_sense = MinSense, parent_formulation = nothing)
> > > ```
> > 
> > 
> > Yes, that makes sense to me. Should the method be kept in "formulation.jl" even with the `::Problem` parameter?
> 
> Good question because we'll have cross-dep issue . Besides I would like to move problem outside MathProg (cf #439) so it's definitely not a good idea.
> EDIT : also confict with #268
> 
> It's also does not look good to have Reformulation as first parameter.
> 
> Do you think we can move the formulation counter into the environment and then have :
> 
> ```julia
> create_formulation!(env::Env, duty::Type{<:AbstractFormDuty}; 
>      obj_sense = MinSense, parent_formulation = nothing)
> ```
> 
> ?
> 
> It fits with the def of Env
> 
> > Env is a place to store global information during the whole execution of Coluna. The environment contains the global parameter ...

I see, I tried to do that just now. Env is defined after including MathProg in "Coluna.jl". I tried to include it after Env definition, but `::Counter` is from MathProg.

guimarqu:
It's ok to move `Counter` into `ColunaBase`.

guimarqu:
By the way, do we really need a Counter to maintain the number of formulations in the Coluna environment ? It looks like engineering.

laradicp:
> By the way, do we really need a Counter to maintain the number of formulations in the Coluna environment ? It looks like we did overengineering.

Could the Counter be a way of passing the argument as a mutable value?

guimarqu:
`Env` is a mutable struct, so I think you can replace the counter by creating a field `nb_formulations` of type `Int` in `Env`.

laradicp:
> `Env` is a mutable struct, so I think you can replace the counter by creating a field `nb_formulations` of type `Int` in `Env`.

Sorry, I meant in the `getnewuid` function, used in the Formulation constructor, the argument passed is only `form_counter`. If it is an `Int`, I think its value cannot be changed. But I can put the operation made in that function outside or pass `Env` instead of `Env.form_counter` as the argument.

guimarqu:
Yes you can pass Env as argument.

Guillaume Marques

src/MathProg/formulation.jl
```diff
@@ -12,29 +12,42 @@ end
...
         parent_formulation = nothing,
         obj_sense::Type{<:Coluna.AbstractSense} = MinSense
-    ) where {Duty<:AbstractFormDuty}
+    )
 
-Construct a `Formulation` of duty `Duty` with objective sense `obj_sense` and parent formulation
-`parent_formulation`.
+Construct a `Formulation` of duty `duty` in the environment `env`, with
+parent formulation `parent_formulation` and objective sense `obj_sense`.
 """
```
guimarqu, line 24: I would say "Create a new formulation in the Coluna's environment `env` with duty `duty`, parent formulation `parent_formulation`, and objective sense `obj_sense`.

src/MathProg/formulation.jl
```diff
@@ -85,6 +98,8 @@ getprimalsolcosts(form::Formulation) = form.manager.primal_sol_costs
 getdualsolmatrix(form::Formulation) = form.manager.dual_sols
 getdualsolrhss(form::Formulation) = form.manager.dual_sol_rhss
 
+"Generates a new `uid` to a formulation in `Env` env."
+getnewuid(env::Coluna.Env) = env.form_counter += 1
```
guimarqu, line 102: I think you can delete this method and do the operation in `create_formulation`.

src/MathProg/formulation.jl
```diff
@@ -12,29 +12,42 @@ end
...
+    Formulation{Original}()
+
+Construct a `Formulation` of duty `Original`, when there is no environment.
+"""
+Formulation{Original}() = Formulation{Original}(
+    0, Counter(), Counter(),
+    nothing, NoOptimizer(), FormulationManager(),
+    MinSense, FormulationBuffer()
+)
+
```
guimarqu, line 50: This method should not exist. Do we create the original formulation before instanciating the `Env` ?

src/optimize.jl
```diff
@@ -54,7 +54,9 @@ function optimize!(prob::MathProg.Problem, annotations::Annotations, params::Par
     env = Env(params)
 
     # Apply decomposition
-    reformulate!(prob, annotations, env)
+    if(prob.re_formulation == nothing)
```
guimarqu, line 57: - Style : no parenthesis for if statement.
- You should use `===` to compare against `nothing`.
- Why do you need this condition ?

guimarqu (commented):
Hi,

You have to update the documentation (you can add the reference here : https://github.com/atoptima/Coluna.jl/blob/master/docs/src/dev/todo.md) to fix the error :  
```
┌ Error: 1 docstring not included in the manual:
│ 
│     Coluna.MathProg.create_formulation! :: Tuple{Env, Type{var"#s2"} where var"#s2"<:Coluna.MathProg.AbstractFormDuty}
│ 
│ 
│ These are docstrings in the checked modules (configured with the modules keyword)
│ that are not included in @docs or @autodocs blocks.
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/6vUwN/src/DocChecks.jl:69
[ Info: Populate: populating indices.
ERROR: LoadError: `makedocs` encountered an error. Terminating build
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] runner(#unused#::Type{Documenter.Builder.RenderDocument}, doc::Documenter.Documents.Document)
   @ Documenter.Builder ~/.julia/packages/Documenter/6vUwN/src/Builder.jl:255
 [3] dispatch(#unused#::Type{Documenter.Builder.DocumentPipeline}, x::Documenter.Documents.Document)
   @ Documenter.Utilities.Selectors ~/.julia/packages/Documenter/6vUwN/src/Utilities/Selectors.jl:170
 [4] #2
   @ ~/.julia/packages/Documenter/6vUwN/src/Documenter.jl:247 [inlined]
 [5] cd(f::Documenter.var"#2#3"{Documenter.Documents.Document}, dir::String)
   @ Base.Filesystem ./file.jl:106
 [6] #makedocs#1
   @ ~/.julia/packages/Documenter/6vUwN/src/Documenter.jl:246 [inlined]
 [7] top-level scope
   @ ~/work/Coluna.jl/Coluna.jl/docs/make.jl:3
in expression starting at /home/runner/work/Coluna.jl/Coluna.jl/docs/make.jl:3
Error: Process completed with exit code 1.

File: src/MathProg/formulation.jl (hunk 3 above)
Comment (laradicp) [reply src/MathProg/formulation.jl:50]: I think so. Isn't `Env` instantiated in `optimize!` ("optimize.jl")?

File: src/optimize.jl (hunk 4 above)
Comment (laradicp) [reply src/optimize.jl:57]: Sorry, I mixed the if statement with C/C++. I need this condition because there was an error when trying to optimize twice. Maybe because a new environment is instantiated each call for `optimize!`, and the same doesn't happen to `Problem`.

File: src/MathProg/formulation.jl (hunk 3 above)
Comment (guimarqu) [reply src/MathProg/formulation.jl:50]: Yes, we can instanciate it when we create `Optimizer`, what do you think ?

File: src/optimize.jl (hunk 4 above)
Comment (guimarqu) [reply src/optimize.jl:57]: haha no problem. Ok makes sense. So if we instanciate the `Env` and `Optimizer` at the same time, it will fix this issue and you will be able to remove this condition.

File: src/MathProg/formulation.jl (hunk 3 above)
Comment (laradicp) [reply src/MathProg/formulation.jl:50]: Yes, that makes sense.

File: src/MOIwrapper.jl [hunk 5]
```diff
@@ -32,14 +31,13 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     default_optimizer_builder::Union{Nothing, Function}
 
     feasibility_sense::Bool # Coluna supports only Max or Min.
-
-    kpis::Union{Nothing, Kpis}
+
```
guimarqu, line 34: whitespace

guimarqu (approved):

PR #495, opened 2021-04-03
Custom data for variables and constraints

guimarqu:
WIP to handle custom representation of variables in constraints in Coluna.


guimarqu:
@rrsadykov @artalvpes You can now pass the custom data from the pricing callback using this method : 

https://github.com/atoptima/Coluna.jl/blob/be98af5d0666c5071a68cf453acdad8407aaa5bf/src/MOIcallbacks.jl#L19-L26

rrsadykov:
Thanks, Guillaume. 
I think the only things missing now are 1) possibility to add a custom model as a subproblem 2) possibility to declare a JuMP (or MOI) variable as a subproblem representative (no need to specify which subproblem) 3) possibility to retrieve Coluna variable which corresponds to  a JuMP (or MOI) variable (so that the mapping can be calculated).

src/MathProg/formulation.jl
```diff
@@ -478,35 +494,68 @@ function _setmembers!(form::Formulation, var::Variable, members::ConstrMembershi
...
             end
         end
     end
     return
 end
 
+# interface ==> move ?
+function computecoeff(::Variable, var_custom_data, ::Constraint, constr_custom_data)
+    error("computecoeff not defined for variable with $(typeof(var_custom_data)) & constraint with $(typeof(constr_custom_data)).")
+end
```
guimarqu, line 525: @artalvpes this is the method you must redefine

test/custom_var_cuts_tests.jl
```diff
@@ -0,0 +1,170 @@
...
+    min_items::Int
+end
+
+function Coluna.MathProg.computecoeff(
+    ::Variable, var_custom_data::MyCustomVarData,
+    ::Constraint, constr_custom_data::MyCustomCutData
+)
+    return (var_custom_data.nb_items >= constr_custom_data.min_items) ? 1.0 : 0.0
+end
+MathOptInterface.Utilities.map_indices(
```
rrsadykov, line 29: Why do we need this?

test/custom_var_cuts_tests.jl
```diff
@@ -0,0 +1,170 @@
...
+        )
+
+        model, x, y, dec = build_toy_model(coluna)
+
+        custom_var_is_added = false
+        custom_cut_is_added = false
+
+        function my_pricing_callback(cbdata)
+            if !custom_var_is_added
+                Coluna.MathProg.addcustomvars!(cbdata.form.parent_formulation, MyCustomVarData)
```
rrsadykov, line 80: Why we cannot do this outside the callback?

test/custom_var_cuts_tests.jl
rrsadykov: I mean `map_indices`

guimarqu:
I think this PR is ready for merge, I put all the things to fix in this issue : https://github.com/atoptima/Coluna.jl/issues/536

test/custom_var_cuts_tests.jl
guimarqu: Follow up in https://github.com/atoptima/Coluna.jl/issues/536

test/custom_var_cuts_tests.jl
guimarqu: Follow up in https://github.com/atoptima/Coluna.jl/issues/536

vitornesello (approved):
tests pass on local env. problem is on dependencies

PR #507, opened 2021-04-15
Move storage & records in ColunaBase

guimarqu:
Task for #330

### DIFF
--- a/src/Algorithm/colgenstabilization.jl
+++ b/src/Algorithm/colgenstabilization.jl
@@ -31,7 +31,7 @@ function ColGenStabRecord(master::Formulation, unit::ColGenStabilizationUnit)
     return ColGenStabRecord(alpha, unit.valid_dual_bound, unit.basestabcenter)
 end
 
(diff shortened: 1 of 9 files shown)


src/Algorithm/colgenstabilization.jl
```diff
@@ -31,7 +31,7 @@ function ColGenStabRecord(master::Formulation, unit::ColGenStabilizationUnit)
     return ColGenStabRecord(alpha, unit.valid_dual_bound, unit.basestabcenter)
 end
 
-function restore_from_record!(
+function ColunaBase.restore_from_record!(
```
rrsadykov, line 34: We are obliged to specify `ColunaBase.` or there is a way to avoid that?

guimarqu: No, I'll remove them at the end. I keep them fro now to track what we redefine from ColunaBase and have a clear idea of the "interface" that devs must implement.

rrsadykov (approved):

PR #515, opened 2021-04-24
Fix call to phase 1 in col gen

### LINKED ISSUE #511: Log message for infeasible node
**Describe the bug**
When solving my application with Coluna, some nodes are clearly infeasible (the dual bound value indicates that there are artificial variables in the master solution), but there is no message saying that the node is infeasible. Messages indicating why a node was pruned would help to keep track of the tree.

**To Reproduce**
```console
julia src/CapacitatedLotSizing/scripts/runsingleinstance.jl testSmall
```
Run with the following parameters (disabled primal heuristics):
```julia
coluna = optimizer_with_attributes(
        Coluna.Optimizer,
        "params" => Coluna.Params(
            solver = Coluna.Algorithm.TreeSearchAlgorithm(
                conqueralg = Coluna.Algorithm.ColCutGenConquer(
                    primal_heuristics = Coluna.Algorithm.ParameterisedHeuristic[])
            ) # default BCP
        ),
        "default_optimizer" => Gurobi.Optimizer # GLPK for the master & the subproblems
    )
```

**Expected behavior**
The following is shown in the log:
```console
***************************************************************************************
**** BaB tree node N° 8, parent N° 2, depth 2, 2 open nodes
**** Local DB = 73.5556, global bounds : [ 73.5556 , 79.0000 ], time = 16.09 sec.
**** Branching constraint: v<=0.0
***************************************************************************************
  <it=  1> <et=16.10> <mst= 0.00> <sp= 0.01> <cols= 0> <al= 0.00> <DB= 1385.9057> <mlp= 1385.9057> <PB=79.0000>
[ Info: Dual bound reached primal bound.
Node is already conquered. No children will be generated.
***************************************************************************************
```
There is a message to indicate that the node is conquered, but not because of infeasibility.

Also, it seems that the following message is not associated with the node:
```console
[ Info: Dual bound reached primal bound.
```

**Environment (please complete the following information):**
 - Julia version 1.6.0
 - OS: Windows 10 64bits
 - Coluna version 0.3.7 #master

guimarqu:
fix #511

### DIFF
--- a/src/Algorithm/colgen.jl
+++ b/src/Algorithm/colgen.jl
@@ -117,11 +117,11 @@ function run!(algo::ColumnGeneration, env::Env, data::ReformData, input::Optimiz
     master = getmaster(reform)
     optstate = OptimizationState(master, getoptstate(input), false, false)
 
-    stop = false
-
     set_ph3!(master) # mixed ph1 & ph2
-    stop, _ = cg_main_loop!(algo, env, 3, optstate, data)
+    # we ignore stop because we need to check if the final LP solution has artificial vars
+    _, _ = cg_main_loop!(algo, env, 3, optstate, data)
 
+    stop = false
     restart = true
     while should_do_ph_1(optstate) && restart && !stop
         set_ph1!(master, optstate)
@@ -133,7 +133,6 @@ function run!(algo::ColumnGeneration, env::Env, data::ReformData, input::Optimiz
     end
 
     @logmsg LogLevel(-1) string("ColumnGeneration terminated with status ", getterminationstatus(optstate))
-
     return OptimizationOutput(optstate)
 end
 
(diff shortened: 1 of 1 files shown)


src/Algorithm/colgen.jl
```diff
@@ -117,11 +117,11 @@ function run!(algo::ColumnGeneration, env::Env, data::ReformData, input::Optimiz
     master = getmaster(reform)
     optstate = OptimizationState(master, getoptstate(input), false, false)
 
-    stop = false
-
     set_ph3!(master) # mixed ph1 & ph2
-    stop, _ = cg_main_loop!(algo, env, 3, optstate, data)
+    # we ignore stop because we need to check if the final LP solution has artificial vars
```
rrsadykov, line 121: I am not sure which should force phase 1 if there are artificial variables in the solution. See my comment in issue #511

PR #522, opened 2021-05-04
UnitsUsageDict -> UnitsUsage

guimarqu:
Task of #330 

I commented EmptyRecord and EmptyStorage because there are not used at the moment.


src/Algorithm/branching/varbranching.jl
```diff
@@ -21,10 +21,15 @@ function generate_children(
...
         #(master, BasisUnit) => READ_AND_WRITE) # not yet implemented
     )
+    set_permission!(
+        units_to_restore,
+        getstoragewrapper(master, MasterBranchConstrsUnit),
+        READ_AND_WRITE
+    )
 
     #adding the first branching constraints
     restore_from_records!(units_to_restore, copy_records(parent.recordids))
```
rrsadykov, line 35: `restore_from_records!()` may also take a vector of pairs (model, unit) instead of the vector of triples (only units which will be restored for writing).

src/ColunaBase/storage.jl
```diff
@@ -74,52 +74,36 @@ restore_from_record!(model::AbstractModel, unit::AbstractStorageUnit, record::Ab
     ))    
 
 
-"""
-    EmptyRecord
+# """
```
rrsadykov, line 77: We need `EmptyRecord` for storage units which are not changed after initialization.

src/ColunaBase/storage.jl
```diff
@@ -161,20 +145,20 @@ function Base.show(io::IO, rw::RecordWrapper{R}) where {R <: AbstractRecord}
     end
 end
 
-"""
-    EmptyRecordWrapper
-"""
+# """
+#     EmptyRecordWrapper
```
rrsadykov, line 149: The same here. `EmptyRecordWrapper` will be needed for empty records.

rrsadykov (changes requested):

src/ColunaBase/storage.jl
guimarqu: I commented because it is not used in the code nor the tests at the moment

guimarqu:
@rrsadykov Ok I understand your comments but I would to discuss about the permissions and this method first: 

https://github.com/atoptima/Coluna.jl/blob/c6e41a67b6c6ed5e9cabdaa458871aeaf1375ced/src/ColunaBase/storage.jl#L263-L265

I think we can merge this PR in the midtime.

PS: tests pass

rrsadykov (approved):

Issue #205, opened 2019-10-16
MILP & LP optimizers for formulation


guimarqu:
Good solution for issue #182 

Each formulation can have have two optimizers : 
- a default optimizer
- a MILP optimizer

If one needs to solve the formulation with integer variables (e.g. restricted ip master heuristic), the default optimizer must support integer variables or the MILP optimizer will be used.

Issue #182, opened 2019-09-13
Optimizer for master problem requires supporting integer variables


mtanneau:
I am trying to use Tulip as the solver for the master problem.

I tried to run `pricing_callback_tests` with Tulip, i.e. I replaced [this line](https://github.com/atoptima/Coluna.jl/blob/master/test/pricing_callback_tests.jl#L29)
with
```julia
build_master_moi_optimizer() = CL.MoiOptimizer(with_optimizer(Tulip.Optimizer)())
```
and sub-problems are still solved with GLPK.

Solving the root seems to be OK, however while running the Master IP heuristic, I get an error since Tulip doesn't support integer variable.
```julia
GAP with ad-hoc pricing callback : Error During Test at /home/mathieu/GitHub/Coluna.jl/test/pricing_callback_tests.jl:36
  Got exception outside of a @test
  MathOptInterface.UnsupportedConstraint{MathOptInterface.SingleVariable,MathOptInterface.Integer}: `MathOptInterface.SingleVariable`-in-`MathOptInterface.Integer` constraint is not supported by the model.
  Stacktrace:
   [1] #correct_throw_add_constraint_error_fallback#31(::MathOptInterface.AddConstraintNotAllowed{MathOptInterface.SingleVariable,MathOptInterface.Integer}, ::Function, ::Tulip.Optimizer{Float64}, ::MathOptInterface.SingleVariable, ::MathOptInterface.Integer) at /home/mathieu/.julia/packages/MathOptInterface/016Ol/src/constraints.jl:151
   [2] correct_throw_add_constraint_error_fallback(::Tulip.Optimizer{Float64}, ::MathOptInterface.SingleVariable, ::MathOptInterface.Integer) at /home/mathieu/.julia/packages/MathOptInterface/016Ol/src/constraints.jl:148
   [3] #throw_add_constraint_error_fallback#27(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Tulip.Optimizer{Float64}, ::MathOptInterface.SingleVariable, ::MathOptInterface.Integer) at /home/mathieu/.julia/packages/MathOptInterface/016Ol/src/constraints.jl:122
[...]

PR #525, opened 2021-05-10
Implementation of column generation stages 

rrsadykov:
Reimplements pull request #519 according to the discussion there


src/Algorithm/pricing.jl
```diff
@@ -0,0 +1,76 @@
+####################################################################
+#                      PricingCallback
+####################################################################
+
+@with_kw struct PricingCallback <: AbstractOptimizationAlgorithm
```
guimarqu, line 5: Can you move this algorithm in `src/Algorithm/basic` ?

src/Algorithm/pricing.jl
```diff
@@ -0,0 +1,76 @@
...
+    stagenumber::Int = 1
```
guimarqu, line 6:
```suggestion
    stage::Int = 1 
```
I suggest to change the name

src/Algorithm/pricing.jl
```diff
@@ -0,0 +1,76 @@
...
+
+####################################################################
+#                      PricingAlgorithm
+####################################################################
+
+@with_kw struct PricingAlgorithm <: AbstractOptimizationAlgorithm
+    callbackalg::PricingCallback = PricingCallback()
+    MIPalg::SolveIpForm = SolveIpForm(deactivate_artificial_vars=false, enforce_integrality=false, log_level=2)
+    dispatch::Int = 0 # 0 - automatic, 1 - impose pricing callback, 2 - impose pricing by MIP 
+end
```
guimarqu, line 56:
```suggestion
@with_kw struct PricingAlgorithm <: AbstractOptimizationAlgorithm
   algorithm = SolveIpForm(deactivate_artificial_vars=false, enforce_integrality=false, log_level=2)
end
```
I think we can have only one parameter which is the algorithm that will be run to solve the pricing subproblem.

Then, we can retrieve the good optimizer, get child algorithms, and running the algorithm by dispatch on the type of `algo.algorithm` : 

```julia
get_child_algorithms(algo::PricingAlgorithm, spform::Formulation{DwSp}) = [(algo.algorithm, spform)]

getoptimizer(spform, ::SolveIpForm) = getmoioptimizer(spform)
getoptimizer(spform, ::PricingCallback) = getuseroptimizer(spform)

# in run!(algo, env, spform, input)

if getoptimizer(spform, algo.algorithm) !== nothing
    # run! sp
else
    # error
end
```

src/MathProg/formulation.jl
```diff
@@ -559,13 +562,20 @@ function constraint_primal(primalsol::PrimalSolution, constrid::ConstrId)
...
-    _initialize_optimizer!(opt, form)
+    form.moioptimizer = opt
+    _initialize_moioptimizer!(opt, form)
     return
 end
 
+function initialize_useroptimizer!(form::Formulation, builder::Function)
+    opt = builder()
+    form.useroptimizer = opt
+    return
```
guimarqu, line 575: what if the user optimizer is a moi optimizer ?

src/MathProg/optimizerwrappers.jl
```diff
@@ -18,11 +18,12 @@ end
 
 struct PricingCallbackData
     form::Formulation
+    stagenumber::Int
     primal_solutions::Vector{PrimalSolution}
 end
 
-function PricingCallbackData(form::F) where {F<:Formulation} 
-    return PricingCallbackData(form, PrimalSolution{F}[])
+function PricingCallbackData(form::F, stagenumber::Int) where {F<:Formulation}
```
guimarqu, line 25: I think we should just use `stage`.

src/Algorithm/colgen.jl
```diff
@@ -25,12 +22,9 @@ restricted master and `pricing_prob_solve_alg` to solve the subproblems.
...
     # TODO : pricing problem solver may be different depending on the
     #       pricing subproblem
-    pricing_prob_solve_alg = SolveIpForm(
-        deactivate_artificial_vars=false,
-        enforce_integrality=false,
-        log_level=2
-    )
+    pricing_prob_solve_alg = PricingAlgorithm()
     essential_cut_gen_alg = CutCallbacks(call_robust_facultative=false)
+    run_only_phase3::Bool = false
```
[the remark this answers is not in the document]
guimarqu: I understood with the example in the test. I think you should rename this parameter `prove_infeasibility` and it should be set to true by conquer for the last stage run.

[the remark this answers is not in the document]
rrsadykov: May be we call this parameter `heuristic_phase`? The idea here is that we should not bother going to phase 1 as there are some other column generation stages remain.

src/Algorithm/conquer.jl
```diff
@@ -165,83 +175,103 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
-                node_pruned = false
-            else # only facultative cut generated
-                node_pruned = node_pruned_by_colgen
-                nb_tightening_rounds += 1
+    nb_cut_rounds = 0
+    stop_conquer = false
+    run_colgen = true
+    while !stop_conquer && run_colgen
+
+        for (stagenumber, colgen) in Iterators.reverse(enumerate(algo.stages))
```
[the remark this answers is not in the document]
rrsadykov: The ideas is that the first stage is always exact (unless the whole B&P is heuristic), and other stages are lighter and lighter heuristics. And we go from stages with higher number to stages with lower number. Of course, we can replace the vector of `ColumnGeneration` by the vector of pairs (stagenumber, `ColumnGeneration`), but this will complicate the parameterisation. What do yo prefer?

src/Algorithm/conquer.jl
```diff
@@ -165,83 +175,103 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
+                colgen_state = getoptstate(colgen_output)
+                update_all_ip_primal_solutions!(nodestate, colgen_state)
+                if nb_lp_primal_sols(colgen_state) > 0
+                    set_lp_primal_sol!(nodestate, get_best_lp_primal_sol(colgen_state))
+                end        
+                status = getterminationstatus(getoptstate(colgen_output))
+                status == TIME_LIMIT && setterminationstatus!(nodestate, status)
+                stop_conquer = getterminationstatus(nodestate) == TIME_LIMIT ||
+                               ip_gap_closed(nodestate, atol = algo.opt_atol, rtol = algo.opt_rtol)
+            end
```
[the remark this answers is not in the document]
rrsadykov: What is `stop_debug`? The idea here is that heuristic stage may improve the best IP solution and the conquer can be stopped by the closure of the IP gap.  Also, the overall time limit may be reached.

src/Algorithm/conquer.jl
```diff
@@ -165,83 +175,103 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
+            stagenumber == algo.minstagenumber && break
```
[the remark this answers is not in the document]
rrsadykov: I agree, i will remove it.

src/Algorithm/conquer.jl
```diff
@@ -165,83 +175,103 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
+                stop_conquer = getterminationstatus(nodestate) == TIME_LIMIT ||
+                               ip_gap_closed(nodestate, atol = algo.opt_atol, rtol = algo.opt_rtol)
+            end
+            stagenumber == algo.minstagenumber && break    
+        end
+    
+        cuts_were_added = false
+        sol = get_best_lp_primal_sol(nodestate)
+        if sol === nothing 
+            @warn "Column generation did not produce an LP primal solution."
```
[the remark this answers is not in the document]
rrsadykov: Ok, if you want

src/Algorithm/conquer.jl
```diff
@@ -165,83 +175,103 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
+                    if cutcb_output.nb_cuts_added == 0
+                        update_ip_primal_sol!(nodestate, sol)
+                    end
                 end
             end
+            ismanager(heur_algorithm) && ColunaBase.restore_from_records!(input.units_to_restore, recordids)
+
+            stop_conquer = getterminationstatus(nodestate) == TIME_LIMIT ||
+                           ip_gap_closed(nodestate, atol = algo.opt_atol, rtol = algo.opt_rtol) 
+            stop_conquer && break
```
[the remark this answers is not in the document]
rrsadykov: ok

src/MathProg/optimizerwrappers.jl
rrsadykov: ok

src/Algorithm/conquer.jl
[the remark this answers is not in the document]
guimarqu: `stop_conquer` sorry.  The stopping criterion is ok but it's a duplication of line 191.

src/Algorithm/conquer.jl
```diff
@@ -165,83 +175,103 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
-            for sol in sort(ip_primal_sols, rev = getobjsense(reform) == MinSense)
-                cutgen = CutCallbacks(call_robust_facultative = false)
-                cutcb_output = run!(cutgen, env, getmaster(reform), CutCallbacksInput(sol))
-                if cutcb_output.nb_cuts_added == 0
-                    update_ip_primal_sol!(nodestate, sol)
+        sort!(heuristics_to_run, by = x -> last(x), rev=true)
+    
+        for (heur_algorithm, name, priority) in heuristics_to_run    
+            @info "Running $name heuristic"
+            ismanager(heur_algorithm) && (recordids = store_records!(reform))
```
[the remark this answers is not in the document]
guimarqu: Yes you're right, I didn't see they were used after the call to `run!`.  I think we should have a `if` block when we assign something.

src/Algorithm/pricing.jl
guimarqu: Ah yes, you are right, this is the problem I was talking about in the previous conversation haha... I need to take an hour to think about this new algorithm.

src/Algorithm/colgen.jl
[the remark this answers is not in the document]
guimarqu: Should we use `stage` if it's something that is not related the column generation algorithm ? What is wrong with a name like `prove_infeasibility`  or `always_prove_infeasibility` ? That's exactly what we do when we perform phase 1 + phase 2.  Otherwise, I prefer to keep `run_only_phase3`.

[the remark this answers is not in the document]
rrsadykov: I will remove this parameter. I think that running or not phase 1 should depend on the dual bound. I propose to run phase 1 if there are art. variables and the solution and column generation converged (if gap is below the optimality tolerance). If the pricing algorithm is a heuristic, it should not set the dual bound, and thus column generation would not update its dual bound, and phase 1 will not be run. I will do a pull for that.

src/Algorithm/conquer.jl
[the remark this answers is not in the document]
rrsadykov: The first variant is implemented (from the exact one to the most heuristic). But I am not sure. Which variant do you suggest? Also, the current convention is that stage 1 in the pricing callback corresponds to the exact stage (so that the dual bound is set equal to primal bound in `PricingCallback` algorithm). Do we keep this convention?

src/Algorithm/conquer.jl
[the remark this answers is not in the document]
rrsadykov: What do you suggest?

src/Algorithm/conquer.jl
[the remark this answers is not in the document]
rrsadykov: ok.

src/MathProg/formulation.jl
rrsadykov: ok

src/Algorithm/pricing.jl
rrsadykov: This is for the case we need both: when heuristic pricing is done in callback and exact pricing is done by MIP. This is quite common to do that.

rrsadykov:
I have pulled the changes requested by Guillaume. 

In the future, the user should provide himself the primal and dual bounds in the callback. I will create an issue for that.

rrsadykov:
I have simplified `ColCutGenConquer` algorithm. In fact, it does not need to know which stage is heuristic or not. This information is passed by the column generation algorithm through bounds.

src/Algorithm/pricing.jl
```diff
@@ -0,0 +1,25 @@
+@with_kw struct DefaultPricing <: AbstractOptimizationAlgorithm
```
guimarqu, line 1: This algorithm should be renamed because it can be used to solve a Formulation outside the column generation algorithm. When we don't know if the optimizer of a formulation is a pricing callback or a MOI optimizer, we'll have to call this method to optimize the formulation (that's why SolveIpForm was supporting both the pricing callback and the MOI optimizer). The name of the algorithm should state that it's going to solve the formulation.

By the way, docstring is missing.

src/decomposition.jl
```diff
@@ -471,7 +475,12 @@ function buildformulations!(
     assign_orig_vars_constrs!(spform, origform, env, annotations, ann)
     create_side_vars_constrs!(spform, origform, env, annotations)
     closefillmode!(getcoefmatrix(spform))
-    initialize_optimizer!(spform, getoptbuilder(prob, ann))
+    initialize_moioptimizer!(spform, getmoioptbuilder(prob, ann))
```
guimarqu, line 478: If the user can use both the pricing callback and the MOI optimizer, what happens if the user didn't specify the constraints of the subproblems ? Are we going to loose time updating the solver buffer that we will never flush ?

src/Algorithm/pricing.jl
```diff
@@ -0,0 +1,25 @@
...
+    callbackalg::PricingCallback = PricingCallback()
```
guimarqu, line 2:
```suggestion
    pricing_callback::PricingCallback = PricingCallback()
```

src/Algorithm/pricing.jl
```diff
@@ -0,0 +1,25 @@
...
+    MIPalg::SolveIpForm = SolveIpForm(deactivate_artificial_vars=false, enforce_integrality=false, log_level=2)
```
guimarqu, line 3:
```suggestion
    solve_ip_form::SolveIpForm = SolveIpForm(deactivate_artificial_vars=false, enforce_integrality=false, log_level=2)
```

src/MathProg/formulation.jl
```diff
@@ -3,7 +3,8 @@ mutable struct Formulation{Duty <: AbstractFormDuty}  <: AbstractFormulation
     var_counter::Int
     constr_counter::Int
     parent_formulation::Union{AbstractFormulation, Nothing} # master for sp, reformulation for master
-    optimizer::AbstractOptimizer
+    moioptimizer::AbstractOptimizer
+    useroptimizer::AbstractOptimizer
```
guimarqu, line 7:
```suggestion
    useroptimizer::Union{Nothing, UserOptimizer}
```
to make sure you receive a UserOptimizer

src/decomposition.jl
```diff
@@ -425,16 +425,20 @@ function assign_orig_vars_constrs!(
...
-        return () -> UserOptimizer(BD.getpricingoracle(ann))
-    end
+function getmoioptbuilder(prob::Problem, ann::BD.Annotation)
     if BD.getoptimizerbuilder(ann) !== nothing
         return () -> MoiOptimizer(BD.getoptimizerbuilder(ann))
     end
     return prob.default_optimizer_builder
 end
 
+function getuseroptbuilder(prob::Problem, ann::BD.Annotation)
```
guimarqu, line 435: is this method called ?

guimarqu (commented):
I'm ok for the new algorithm that call either PricingCallback or the MOI optimzer. However, I'm afraid that the MOI optimizer is always provided even if the formulation is not entered by the user.

src/Algorithm/pricing.jl
guimarqu: Ok so in this PR you don't just implement column generation stages. You extend the use of the pricing callback and you suggest a first solution to #182 and #205.

src/Algorithm/pricing.jl
rrsadykov: I do not completely understand why this algorithm can be called outside column generation. It uses **pricing** callback. What is the other use case?

src/Algorithm/conquer.jl
```diff
@@ -117,37 +117,45 @@ end
...
         preprocess = PreprocessAlgorithm()
         cutgen = CutCallbacks()
         run_preprocessing::Bool = false
     )
 
     Column-and-cut-generation based algorithm to find primal and dual bounds for a 
-    problem decomposed using Dantzig-Wolfe paradigm. It applies `colgen` for the column 
-    generation phase, `cutgen` for the cut generation phase, and it can apply several primal
+    problem decomposed using Dantzig-Wolfe paradigm. It applies `stages` of the column generation 
+    algorithm (for example, the exact stage and the heuristic stage), `cutgen` for the cut
```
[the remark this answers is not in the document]
rrsadykov: ok

src/decomposition.jl
rrsadykov: I did not think about this. Is it a big overhead if we have MOI optimizer for the subproblem we never use?

src/Algorithm/pricing.jl
rrsadykov: ok

src/Algorithm/pricing.jl
rrsadykov: ok

src/decomposition.jl
rrsadykov: deleted

src/MathProg/formulation.jl
```diff
@@ -3,7 +3,8 @@ mutable struct Formulation{Duty <: AbstractFormDuty}  <: AbstractFormulation
     var_counter::Int
     constr_counter::Int
     parent_formulation::Union{AbstractFormulation, Nothing} # master for sp, reformulation for master
-    optimizer::AbstractOptimizer
+    moioptimizer::AbstractOptimizer
```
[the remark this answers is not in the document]
rrsadykov: ok

src/MathProg/formulation.jl
rrsadykov: ok

src/MathProg/formulation.jl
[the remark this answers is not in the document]
rrsadykov: I cannot do that because `MoiOptimizer` is defined later in optimizerwrappers.jl. And I cannot move   optimizerwrappers.jl above formulation.jl, as `Formulation` is user there.

src/MathProg/formulation.jl
rrsadykov: Same here

src/Algorithm/pricing.jl
guimarqu: It is more than a pricing callback, it's the solver of the subproblem. Pricing callback is a use case of this solver.

```julia
BD.specify!.(subproblems, lower_multiplicity = 0, solver = my_pricing_callback)
```

src/decomposition.jl
guimarqu: I don't know but we should avoid updating the MOI optimizer if the user does not provide the whole formulation.

src/MathProg/formulation.jl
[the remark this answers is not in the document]
guimarqu: Ok I see. That's a problem.

src/decomposition.jl
rrsadykov: I do not know how to do it.

src/Algorithm/pricing.jl
rrsadykov: So, which name do you suggest?

rrsadykov:
Guillaume, may be you implement yourself the changes you want to do? There should remain few things to do.

artalvpes:
Hi @rrsadykov ,

    I just started looking at this version. The interface is really much more intuitive! Congratulations! However, there were two features in the original version that are missing here, which I consider very important: showing the current pricing stage in the log file, and counting the number of of calls to the exact stage in the test. The former helps to check whether the performance of the solver is as expected when solving larger instances, and the latter is to make sure that the heuristic pricing is effective to save some exact pricing calls. For example, the heuristic princing might never be called or might never produce useful columns or such columns might never be inserted in the master and the test would pass (relying only in the exact phase to solve the relaxations).

guimarqu:
I suggest to keep the name `SolveIpForm` for the optimizer of the formulation with integer variables.
We keep the algorithm split `MoiSolver` / `UserSolver`. I like it, it's cleaner.
`SolveIpForm` automatically calls the appropriate algorithm `MoiSolver` / `UserSolver` depending on the type of the solver.

We can change the `specify!` method in BlockDecomposition to allow the user to pass a vector of optimizers : 

```julia
BD.specify!.(subproblems, lower_multiplicity = 0, solver = [MOI.Gurobi, my_callback_stage2, my_callback_stage3])
```

and we change the `dispatch` argument into `solver` : 

```julia
stages = [ 
    ClA.ColumnGeneration(pricing_prob_solve_alg = ClA.SolveIpForm(solver = 1)),
    ClA.ColumnGeneration(pricing_prob_solve_alg = ClA.SolveIpForm(solver = 2)),
    ClA.ColumnGeneration(pricing_prob_solve_alg = ClA.SolveIpForm(solver = 3)),
]
```

So in that case, we do the exact phase with `MOI.Gurobi` and we use a callback for stages 2 & 3. The user will have to write a callback for each phase.

I don't see any better solution at the moment to handle https://github.com/atoptima/Coluna.jl/pull/525#discussion_r631094863.

Moreover, if we don't fix this, we'll have a design issue with customized optimizers. I would like to have these optimizers connected to Coluna through MOI. So to avoid buffering changes several times, I suggest to identify the solvers that have the same MOI optimizers and set the parameters just before optimizing the formulation. For instance, if you define solvers of your subproblem like this,

```julia
solvers = [
     with_attributes(MOI.RCSP, mode = :exact), 
     with_attributes(MOI.RCSP, mode = :heuristic),
     with_attributes(MOI.RCSP, mode = :speed_heuristic)
]
```

Coluna will instanciate `MOI.RCSP` only once for the subproblem, and will set the parameters just before optimizing the subproblem. We'll have to find a way to set parameters only when needed (because changing parameters may need some calculations ?).

What do you think ? (cc @vitornesello )

rrsadykov:
> So in that case, we do the exact phase with `MOI.Gurobi` and we use a callback for stages 2 & 3. The user will have to write a callback for each phase.

Should the user also write a callback for the  `MOI.Gurobi`  stage?

guimarqu:
> > So in that case, we do the exact phase with `MOI.Gurobi` and we use a callback for stages 2 & 3. The user will have to write a callback for each phase.
> 
> 
> 
> Should the user also write a callback for the  `MOI.Gurobi`  stage? 
> 
> 
> 
> Is it possible here to give attributes like `solver=JuMP.optimizer_with_attributes(GLPK.Optimizer, "tm_lim" => 60 * 2_200)`?

No the user doesn't write a callback when he uses a MOI optimizer. He can pass an optimizer with attributes (already working).

rrsadykov:
If the solver in `SolveIpForm` is fixed, it is not worth anymore for me to define child algorithms `MoiSolver` and `UserSolver`. Instead we may have `SolveIpFormWithMOI` and `SolveIpFormWithUserSolver`. So we may have something like 

```julia
BD.specify!.(subproblems, lower_multiplicity = 0, user_solvers = [my_callback_stage2, my_callback_stage3])
```
and then
```julia
stages = [ClA.ColumnGeneration(pricing_prob_solve_alg = ClA.SolveIpFormWithMOI(moi=MOI.Gurobi)),
          ClA.ColumnGeneration(pricing_prob_solve_alg = ClA.SolveIpFormWithUserSolver(solver = 1)),
          ClA.ColumnGeneration(pricing_prob_solve_alg = ClA.SolveIpFormWithUserSolver(solver = 2))
          ]
```

@guimarqu, what do you think?

rrsadykov:
Also, I do not understand how this would allow us to avoid the problem raised in [#525 (comment)](https://github.com/atoptima/Coluna.jl/pull/525#discussion_r631094863). How Coluna knows whether to create `moioptimizer` or not?

guimarqu:
We can keep the split even if the user won't call these two algorithms.

Your solution won't work if one wants to optimize au problems with different solvers.

If the solver builder stored in the annotation of the subproblem returns a MOI optimizer, Coluna will create a MoiOptimizer. Otherwise it's a UserOptimizer.

rrsadykov:
> I would like to have customized optimizers connected to Coluna through MOI.

I do not think this is good idea. First, custom optimizer is not necessarily a mathematical optimizaton algorithm to be connected through MOI (Mathematical Optimization Interface). For example, RCSP solver is not a mathematical optimization solver.  Second, it does not suffice to write `solvers = [with_attributes(MOI.RCSP, mode = :exact)]` to use RCSP solver for the pricing. You need also to define your graph, resources, etc. Moreover, you need to connect also additional algorithms, like reduced cost fixing, enumeration, also you need to define additional storage units. 

We really like the user interface we have in VrpSolver (through mapping). For me, Coluna should be able to allow dependent packages to use any user interface they want to define subproblems (through custom annotations?).

guimarqu:
> > I would like to have customized optimizers connected to Coluna through MOI.
> 
> 
> 
> I do not think this is good idea. First, custom optimizer is not necessarily a mathematical optimizaton algorithm to be connected through MOI (Mathematical Optimization Interface). For example, RCSP solver is not a mathematical optimization solver.  Second, it does not suffice to write `solvers = [with_attributes(MOI.RCSP, mode = :exact)]` to use RCSP solver for the pricing. You need also to define your graph, resources, etc. Moreover, you need to connect also additional algorithms, like reduced cost fixing, enumeration, also you need to define additional storage units. 
> 
> 
> 
> We really like the user interface we have in VrpSolver (through mapping). For me, Coluna should be able to allow dependent packages to use any user interface they want to define subproblems (through custom annotations?).

Yes I agree. I was speaking only about the communication between Coluna and the solver. For instance, when Coluna changes the cost of the subproblem variables to give the reduced costs to the subsolver or also the bounds of the variables.

rrsadykov:
> Yes I agree. I was speaking only about the communication between Coluna and the solver. For instance, when Coluna changes the cost of the subproblem variables to give the reduced costs to the subsolver or also the bounds of the variables.

I think Coluna just needs to calculate the reduced costs (and bounds, etc) of the representative variables in the master (`MasterRepPricingVar`). This is sufficient. Then pricing solver will use this information to run (+ dual values of non-robust cuts), and then it returns the solution which is defined on these representative variables (+ custom info if needed).

I think that `BD.specify!()` is not needed for custom pricing solvers. `BlockDecomposition` should only provide interface for 1) adding custom models as subproblems and 2) declaring which variables defined by the user are sp. representative variables and which variables are standard (pure master). No variables need to be defined in custom subproblems.

rrsadykov:
> However, there were two features in the original version that are missing here, which I consider very important: showing the current pricing stage in the log file, and counting the number of of calls to the exact stage in the test.

Current pricing stage is shown in `run!(algo::ColCutGenConquer,...)` before running this stage: 

```
            if length(algo.stages) > 1 
                @logmsg LogLevel(0) "Column generation stage $stage is started"
            end
```

Counting the number of calls can easily be counted in the callback.

guimarqu:
> > Yes I agree. I was speaking only about the communication between Coluna and the solver. For instance, when Coluna changes the cost of the subproblem variables to give the reduced costs to the subsolver or also the bounds of the variables.
> 
> I think Coluna just needs to calculate the reduced costs (and bounds, etc) of the representative variables in the master (`MasterRepPricingVar`). This is sufficient. Then pricing solver will use this information to run (+ dual values of non-robust cuts), and then it returns the solution which is defined on these representative variables (+ custom info if needed).
> 
> I think that `BD.specify!()` is not needed for custom pricing solvers. `BlockDecomposition` should only provide interface for 1) adding custom models as subproblems and 2) declaring which variables defined by the user are sp. representative variables and which variables are standard (pure master). No variables need to be defined in custom subproblems.

`BD.specify!` is needed for the custom pricing solvers. That's how you pass the information to Coluna. 1) the custom model will be given together with the solver. 2) No variable will be defined in the custom subproblems but when the user writes the subproblems models, he will have to map the subproblem variables to the decisions taken in the subproblem. This is specific to the custom solver.

If you agree with this future change https://github.com/atoptima/Coluna.jl/pull/525#issuecomment-841573504, I think it's ready for merge.

rrsadykov:
> `BD.specify!` is needed for the custom pricing solvers. That's how you pass the information to Coluna. 1) the custom model will be given together with the solver. 2) No variable will be defined in the custom subproblems but when the user writes the subproblems models, he will have to map the subproblem variables to the decisions taken in the subproblem. This is specific to the custom solver.

Ok, custom models can be given with `BD.specify!`.  The pricing solver can be given by directly setting `pricing_prob_solve_alg` in `ColumnGeneration`. Yes, mapping is the responsibility of the custom subproblem. It just needs to have access to subproblem representative variables in the master.

rrsadykov:
> If you agree with this future change [#525 (comment)](https://github.com/atoptima/Coluna.jl/pull/525#issuecomment-841573504), I think it's ready for merge.

Ok for me.

guimarqu (approved):

PR #528, opened 2021-05-15
Support of variable branching priorities in the branching algorithm

rrsadykov:
This functionality works, but we also need to change the type of branching priorities in `BlockDecomposition` from `Int` to `Float`.  

The idea of fractional branching priorities to have both "soft" and "hard" branching priorities in strong branching:
- If the number of branching candidates with priority 4.0 is less than the maximum number, no branching candidates with priority 3.0 will be considered. 
- If the number of branching candidates with priority 3.5 is less than the maximum number, then some branching candidates with priority 3.0 will be considered (to bring the total number to the maximum). 
- If the number of branching candidates with priority 3.5 is not less than the maximum number, then no branching candidates with priority 3.0 will be considered.

### DIFF
--- a/docs/src/dev/formulation.md
+++ b/docs/src/dev/formulation.md
@@ -77,6 +77,7 @@ activate!
 deactivate!
 isexplicit
 getname
+getbranchingpriority
 ```
 
 ```@meta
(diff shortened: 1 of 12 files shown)


guimarqu (approved):
Ok I'll change the branching priority type in BlockDecomposition. I didn't know this mechanism.

PR #534, opened 2021-05-23
Vector of optimizers in `Formulation`

guimarqu:
~~- needs Manifest because it uses master branch of BlockDecomposition~~

Design changes : 
- Formulation has a vector of optimizers (MoiOptimizer, PricingCallback or CustomOptimizer later)
- ~~run! methods of `SolveIpForm`, `SolveLpForm`, and `PricingCallback` receive a new optional argument `optimizer_id` which is the index of the optimizers attached to the Formulation.~~ See https://github.com/atoptima/Coluna.jl/pull/534#issuecomment-848047440
- methods to synchronize formulation with MoiOptimizers receive the optimizer as argument


rrsadykov:
I do not like having an optional argument in the `run!()` method. If we change the algorithm for another optimization algorithm without this argument in the `run!()` method, we would have en error. The purpose of algorithms to be interchangeable. If we cannot change an algorithm for another, it is not an algorithm anymore. 

Why not put `optimizer_id` directly to the definition of `SolveIpForm`, `SolveLpForm`, and `PricingCallback`? This would simplify a lot of things, and with this we would avoid adding arguments to the `run!()` method.

Also, the purpose of `DefaultPricing` algorithm is not clear anymore. The purpose of this algorithm was to switch pricing algorithms in the runtime. As now the user needs to set `optimizer_id`, we do not need to switch algorithms in the runtime. Why not setting directly `SolveIpForm` or `PricingCallback` with right `optimizer_id` as `pricing_prob_solve_alg` of `ColumnGeneration`?

guimarqu:
> I do not like having an optional argument in the `run!()` method. If we change the algorithm for another optimization algorithm without this argument in the `run!()` method, we would have en error. The purpose of algorithms to be interchangeable. If we cannot change an algorithm for another, it is not an algorithm anymore.

Since the argument is optional, you'll not have an error but you always have to use the first optimizer which is for me a major drawback. I tried to follow what I said in the discussion in your PR and to be quite conservative before having discussion.

> 
> Why not put `optimizer_id` directly to the definition of `SolveIpForm`, `SolveLpForm`, and `PricingCallback`? This would simplify a lot of things, and with this we would avoid adding arguments to the `run!()` method.
> 
> Also, the purpose of `DefaultPricing` algorithm is not clear anymore. The purpose of this algorithm was to switch pricing algorithms in the runtime. As now the user needs to set `optimizer_id`, we do not need to switch algorithms in the runtime. Why not setting directly `SolveIpForm` or `PricingCallback` with right `optimizer_id` as `pricing_prob_solve_alg` of `ColumnGeneration`?

This is the two options I saw this afternoon when working on this PR : 
- if we remove `DefaultPricing`, keep  `SolveIpForm` `SolveLpForm` `PricingCallback` & we put  `optimizer_id` as parameter of these algorithms,  then Coluna will throw an error when the user choses `SolveLpForm` to call a `UserOptimizer`. I think we should avoid this by automatically detecting if the solver is a `MoiOptimizer` or a `UserOptimizer` (that's the job of `DefaultPricing`, it should be renamed by the way).
- if we keep  `SolveIpForm` `SolveLpForm` `PricingCallback`, put  `optimizer_id` as parameter of these algorithms & let`DefaultPricing` do the dispatch depending on the type of the optimizer, then we have to make the types `SolveIpForm`, `SolveLpForm`,`PricingCallback` mutable to allow `DefaultPricing` to change their `optimizer_id` parameter before calling them. I don't know if it's ok to have mutable parameters for the algorithms.

What do you think ?

rrsadykov:
> Since the argument is optional, you'll not have an error but you always have to use the first optimizer which is for me a major drawback. I tried to follow what I said in the discussion in your PR and to be quite conservative before having discussion.

If you have an optional argument, you will use it at least once. In the call when this argument is used, you cannot use another algorithm which does not have this argument in the `run!()` function, you will get an error. Therefore, these two algorithms, one which calls and another which is called, are not independent anymore. Thus, you need to combine these algorithms together (or you should introduce another algorithm type, and add this argument to the input). 

> Coluna will throw an error when the user chooses `SolveLpForm` to call a `UserOptimizer`

I do not see a problem here. If parameterisation is inconsistent, there should be an error, it is normal for me. We cannot expect the user to always give a consistent parameterisation, especially if we mix paramerization in the algorithms and through `BlockDecomposition`. 

I propose to put `optimizer_id` as a parameter in the algorithms with default value 1. We can leave `DefaultPricing` algorithm and rename it if you want. It will call the callback if the optimizer of formulation with id equal to `optimizer_id` of `CallbackAlgorithm` is of type `UserOptimizer`, otherwise it will call `SolveIpForm`. With this, the behaviour would be the same as before if the user gives only one solver for the subproblem. If no solver is set, then Coluna may automatically set the MOI default optimizer. However, if more two or more solves are set, then user needs to use stages in column generation, and this requires a proper parameterisation, which may trigger errors if not properly set.

rrsadykov:
> * I don't know if it's ok to have mutable parameters for the algorithms.

We should avoid that of course

rrsadykov:
Ok, I like the current version better.

src/Algorithm/colgen.jl
```diff
@@ -22,7 +22,7 @@ restricted master and `pricing_prob_solve_alg` to solve the subproblems.
     restr_master_optimizer_id = 1
     # TODO : pricing problem solver may be different depending on the
     #       pricing subproblem
-    pricing_prob_solve_alg = DefaultPricing()
+    pricing_prob_solve_alg = SolveIpForm()
```
rrsadykov, line 25: I think you forgot

```
SolveIpForm(
    moi_params = MoiOptimize(deactivate_artificial_vars=false,
                             enforce_integrality=false,
                             log_level=2)
)

```

test/pricing_callback_tests.jl
```diff
@@ -10,12 +10,12 @@ function pricing_callback_tests()
                 solver = ClA.TreeSearchAlgorithm(
                     conqueralg = ClA.ColCutGenConquer(
                         stages = [ClA.ColumnGeneration(
-                                    pricing_prob_solve_alg = ClA.DefaultPricing(
-                                        optimizer_id=2, pricing_callback = PricingCallback(stage=1) 
+                                    pricing_prob_solve_alg = ClA.SolveIpForm(
+                                        optimizer_id=2, user_params = ClA.UserOptimize(stage=1)
```
rrsadykov, line 14: If user may define several solvers, he can define several callbacks (one heuristic and second exact). This way we can avoid `stage` parameter: user will just define `optimizer_id=2` and `optimizer_id=3`. @guimarqu, what do you think about this?

guimarqu: I thought about that but forgot. So it's ok for me.

guimarqu:
> Ok, I like the current version better.

I was planning to write a comment to explain the changes after my break.

`MoiOptimize` and `UserOptimize`  are not classic algorithm anymore although you can use them as algorithm. The user should rather use `SolveIpForm`. It's kind of halfway between the implementation of `SolveIpForm` in master & what you did in the previous PR.

I'm going to fix your comments.

test/pricing_callback_tests.jl
guimarqu: I open a new issue to adapt this

src/Algorithm/basic/solveipform.jl
```diff
@@ -1,29 +1,114 @@
...
+_optimizer_params(algo::SolveIpForm, ::UserOptimizer) = algo.user_params
+# TODO : custom optimizer
+_optimizer_params(::SolveIpForm, ::NoOptimizer) = nothing
+
+function run!(algo::SolveIpForm, env::Env, form::Formulation, input::OptimizationInput)::OptimizationOutput
+    opt = getoptimizer(form, algo.optimizer_id)
+    params = _optimizer_params(algo, opt)
+    if params !== nothing
+        return run!(params, env, form, input; optimizer_id = algo.optimizer_id)
+    end
```
vitornesello, line 90: I think that there is some dispatch happening in runtime on lines 86, 87 and 89.
I think we should open an issue to investigate the impact later

vitornesello (approved):

rrsadykov:
Ok for me, but I think the tests should pass before approving.

guimarqu:
Tests pass on my machine. These changes need BlockDecomposition#master & when I push the Manifest, it breaks the CI. I don't want to release a new version of BlockDecomposition before having custom model/solver.

PR #535, opened 2021-05-26
Prototyping custom model/optimizer

guimarqu:
First implementation for the custom model/optimizer.

At the moment it's not possible to replace the `Formulation` by a custom one because it requires much more work.


guimarqu:
@laradicp I wasn't working when direct mode was deactivated, so I made some modifications.

However, I don't think it's a good solution because I think we should build the mapping `custom model decision -> VarId` once (at the beginning).

It's more difficult than I thought. We must figure out how we can build this relation `custom model decision -> VarId` knowing that the user can only give a relation between `custom model decision` and `JuMP.VariableRef`. 

My first idea was to use `MOI.VariableIndex -> VarId` from `Optimizer.varids` but we don't have access to the optimizer in the `run!` method of the custom optimizer. Therefore, I tried to move `Optimizer.varids` into `Env`, but it breaks everything (circular dep `Env` & `MathProg`).

test/interfaces/model.jl
```diff
@@ -7,23 +7,28 @@ mutable struct KnapsackLibModel <: Coluna.MathProg.AbstractFormulation
...
 setcapacity!(model::KnapsackLibModel, cap) = model.capacity = cap
 setweight!(model::KnapsackLibModel, j::Int, w) = model.weights[j] = w
 setcost!(model::KnapsackLibModel, j::Int, c) = model.costs[j] = c
-pushvarid!(model::KnapsackLibModel, x::VariableRef) = push!(
-    model.varids, x.model.moi_backend.varids[x.index]
-)
-map!(model::KnapsackLibModel, j::Int) = push!(
-    model.map, model.varids[j] => 1.0
-)
+map!(model::KnapsackLibModel, j::Int, x::JuMP.VariableRef) = model.job_to_jumpvar[j] = x
```
laradicp, line 21: So the goal is to have `job_to_varid::Dict{Int, Coluna.MathProg.VarId}`? Why can't we use
 ```julia
map!(model::KnapsackLibModel, j::Int, x::JuMP.VariableRef) = model.job_to_jumpvar[j] = x.model.moi_backend.varids[x.index]
```
here?

Comment (laradicp) [reply test/interfaces/model.jl:21]: Ah because of direct mode, right?

Comment (guimarqu) [reply test/interfaces/model.jl:21]: Yes because of when you're not using the direct mode, there is no `Coluna.Optimizer` when you build the custom model. So it doesn't work :/

Comment (laradicp) [reply test/interfaces/model.jl:21]: Yes, I see... I'll try to think of something else

File: Project.toml [hunk 2]
```diff
@@ -10,6 +10,7 @@ DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 DynamicSparseArrays = "8086fd22-9a0c-46a5-a6c8-6e24676501fe"
+KnapsackLib = "86859df6-51c5-4863-9ac2-2c1ab8e53eb2"
```
guimarqu, line 13:
```suggestion
```

Project.toml
```diff
@@ -18,7 +19,6 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 
 [compat]
-BlockDecomposition = "~1.3"
 DataStructures = "0.17, 0.18"
```
guimarqu, line 22:
```suggestion
BlockDecomposition = "~1.3"
DataStructures = "0.17, 0.18"
```

guimarqu:
Tests are ok

test/interfaces/model.jl
```diff
@@ -0,0 +1,136 @@
...
+setcapacity!(model::KnapsackLibModel, cap) = model.capacity = cap
+setweight!(model::KnapsackLibModel, j::Int, w) = model.weights[j] = w
+setcost!(model::KnapsackLibModel, j::Int, c) = model.costs[j] = c
+map!(model::KnapsackLibModel, j::Int, x::JuMP.VariableRef) = model.job_to_jumpvar[j] = x
+
+coluna_backend(model::MOI.Utilities.CachingOptimizer) = coluna_backend(model.optimizer)
+coluna_backend(b::MOI.Bridges.AbstractBridgeOptimizer) = coluna_backend(b.model)
+coluna_backend(model) = model
+
+function get_coluna_varid(model::KnapsackLibModel, form, j::Int)
```
rrsadykov, line 27: It seems this function is not used

test/interfaces/model.jl
```diff
@@ -0,0 +1,136 @@
...
+    jumpvar = model.job_to_jumpvar[j]
+    opt = coluna_backend(backend(jumpvar.model))
+    return Coluna._get_orig_varid_in_form(opt, form, jumpvar.index)
+end
```
guimarqu, line 31:
```suggestion
```

test/interfaces/model.jl
guimarqu: Yes, I deleted it.

rrsadykov:
I find this organization strange. You have subproblem formulation, inside you have optimizer, which itself has `AbstractFormulation` inside (I think that `AbstractFormulation` should be renamed to `AbstractModel`). It would be much simpler to directly have an `AbstractModel` as a subproblem formulation.

In `specify!` instead giving `solver`, we may define directly a model. For me it should be something like

```
mutable struct KnapsackLibModel <: Coluna.AbstractModel
    master::Coluna.MathProg.Formulation
    nbitems::Int
    costs::Vector{Float64}
    weights::Vector{Float64}
    capacity::Float64
    job_to_jumpvar::Dict{Int, JuMP.VariableRef}
end

function setmaster!(knap_model::KnapsackLibModel, master::Coluna.MathProg.Formulation)
    knap_model.master = master
end

struct KnapsackAlgorithm <: Coluna.AbstractOptimisationAlgorithm
end

function Coluna.Algorithm.run!(
    algo::KnapsackAlgorithm, env::Coluna.Env, model::KnapsackLibModel,
    input::Coluna.Algorithm.OptimizationInput)
    costs = -[Coluna.MathProg.getcurcost(model.master, _getvarid(model.master, env, job_to_jump_var[j])) for j in 1:length(opt.model.costs)]
    ...
    for j in selected
        push!(varids, _getvarid(model.master, env, job_to_jump_var[j]))
        push!(varvals, 1)
    end
    ...

end

....

coluna = JuMP.optimizer_with_attributes(
    Coluna.Optimizer,
    "params" => CL.Params(solver = TreeSearchAlgorithm(
        conqueralg=ColCutGenConquer(
            colgen=ColumnGeneration(pricing_prob_solve_alg=KnapsackAlgorithm())
        )
    )),
    "default_optimizer" => GLPK.Optimizer
)

model = BlockModel(coluna; direct_model = true)
@variable(model, x[m in data.machines, j in data.jobs], Bin)
@constraint(model, 
    sp[j in data.jobs], sum(x[m,j] for m in data.machines) == 1
)
@objective(model, Min,
    sum(data.cost[j,m]*x[m,j] for m in data.machines, j in data.jobs)
)

@axis(M, data.machines)
@dantzig_wolfe_decomposition(model, dec, M)

sp = getsubproblems(dec)
for m in M
    knp_model = KnapsackModel(length(data.jobs))
    setcapacity!(knp_model, data.capacity[m])
    for j in data.jobs
        setweight!(knp_model, j, data.weight[j,m])
        setcost!(knp_model, j, data.cost[j,m])
        map!(knp_model, j, x[m,j])
        #here you need to say that x[m,j] is an implicit representative variable
    end
    specify!(sp[m], model = knp_model) 
end

```

We may define `AbstractCustomSpModel <: AbstractModel` for which function `setmaster!` should be defined. So Coluna will call this function if it sees an `AbstractCustomSpModel`.

There is also a functionality missing, which allows us to say that a certain variable is implicit representative.

Concerning setup variables, I think they should should be managed automatically by Coluna

rrsadykov:
Ok, may be sp. setup variables should also be managed by the user. So we may have something like

```
@variable(model, setup[m in data.machines], Int)
...
set_setup_var!(knp_model[m], setup[m])
...
specify!(sp[m], model = knp_model, setupvar = setup[m]) 
```

With the knowledge of setup variable, the user can check whether a certain solution has a negative reduced cost.

guimarqu:
Follow up in https://github.com/atoptima/Coluna.jl/issues/537

rrsadykov:
Ok for me. Why the tests do not pass?

guimarqu:
Because we need to release a new version of BlockDecomposition

rrsadykov (approved):

Issue #536, opened 2021-06-04
Follow up of "custom data"


guimarqu:
- [x] Create a `AbstractCustomData` type to make clear what is the super type of the field `custom_data` in `Variable` & `Constraint` & also the type of `custom_data` in some methods.

- [x] Create methods `addcustomconstrs!(blockmodel, custom_data_type)` & `addcustomvars!(blockmodel, custom_data_type)` in BlockDecomposition & call methods `Coluna.MathProg.addcustomconstrs!` when we create the master formulation. Then, we can remove : 
https://github.com/atoptima/Coluna.jl/blob/36e2d0d8f40ba964637c1d6da8b0683ffd776f08/test/custom_var_cuts_tests.jl#L79-L82  and https://github.com/atoptima/Coluna.jl/blob/36e2d0d8f40ba964637c1d6da8b0683ffd776f08/test/custom_var_cuts_tests.jl#L138-L141

- [x] Move the following method  in BlockDecomposition. Check if we can untype all arguments.
https://github.com/atoptima/Coluna.jl/blob/36e2d0d8f40ba964637c1d6da8b0683ffd776f08/test/custom_var_cuts_tests.jl#L35-L37

- [x] Move the following methods in BlockDecomposition :
https://github.com/atoptima/Coluna.jl/blob/36e2d0d8f40ba964637c1d6da8b0683ffd776f08/test/custom_var_cuts_tests.jl#L29-L34

- [x] Create new methods in MathProg to set/get current `inc_val` and get perene `inc_val` of a constraint. Then, update the following pieces of code : 
https://github.com/atoptima/Coluna.jl/blob/36e2d0d8f40ba964637c1d6da8b0683ffd776f08/test/custom_var_cuts_tests.jl#L94
https://github.com/atoptima/Coluna.jl/blob/36e2d0d8f40ba964637c1d6da8b0683ffd776f08/src/Algorithm/colgen.jl#L367-L373
Hint for docstring : 
> [22:00] Artur Pessoa: A constraint is a variable in the dual. So, it seems natural to me that the incumbent value of the constraint is the dual value associated to it.

- [x] Change the output when we separate cuts with custom data ? 

`Robust cut separation callback adds 0 new essential cuts and 1 new facultative cuts.`

- [x] Documentation

guimarqu:
An example with documentation is available on master.

PR #538, opened 2021-06-07
Follow up of "custom data"

laradicp:


src/decomposition.jl
```diff
@@ -51,6 +53,8 @@ function instantiatemaster!(
         obj_sense = getobjsense(get_original_formulation(prob))
     )
     setmaster!(reform, masterform)
+    Coluna.MathProg.addcustomvars!(reform.master, ColunaBase.AbstractCustomData)
+    Coluna.MathProg.addcustomconstrs!(reform.master, ColunaBase.AbstractCustomData)
```
laradicp, line 57: @guimarqu I didn't understand why we should create these methods in BlockDecomposition if we would keep calling the Coluna.MathProg ones. Is it for another purpose or should I change them to the ones created in BlockDecomposition?

guimarqu: We should create them in BlockDecomposition because we should call them once when we reformulate. So BlockDecomposition must pass the information to Coluna.

src/Algorithm/colgen.jl
```diff
@@ -375,12 +375,12 @@ function solve_sps_to_gencols!(
         updatereducedcosts!(reform, redcostshelper, smooth_dual_sol)
     end
 
-    # udate the incumbent values of constraints
+    # update the incumbent values of constraints
     for (_, constr) in getconstrs(masterform)
-        constr.curdata.inc_val = 0.0
+        MathProg.setcurincval!(masterform, constr, 0.0)
```
guimarqu, line 380: I think you can export this method.

laradicp: I'm not sure I got the right meaning of this, but I've seen now that I duplicated these functions. They already existed and they are already exported.

guimarqu: Ok perfect then. If it's exported you can remove the `MathProg` prefix. If tests pass on your computer, we can merge then.

laradicp: Many things were wrong and I didn't notice because there was an unrelated test failing before `custom_var_cuts_test`, which therefore never ran. Now I fixed some of the bugs but can't get the right result yet, I'll keep trying. Also I had to move `addcustomvars!` and `addcustomconstrs!` to a function where `custom_data` is reachable so it could give the right key to the dictionary, is that ok? I think it's possible to search for the key `AbstractCustomData` instead, if that's better.

src/MathProg/varconstr.jl
```diff
@@ -264,7 +263,8 @@ end
     getperenincval(formulation, varconstr)
 
 Return the incumbent value as defined by the user of a variable or a constraint in a formulation. 
-The incumbent value is ?
+For a variable, the incumbent value is ?
```
guimarqu, line 266: We can say it's a primal value associated to the variable (ping @artalvpes).

guimarqu (approved):

guimarqu:
task for #536

PR #543, opened 2021-06-12
docstring for restricted master heuristic

### LINKED ISSUE #460: missing docstring for `DefaultRestrictedMasterHeuristic`

guimarqu:
fix #460

### DIFF
--- a/src/Algorithm/conquer.jl
+++ b/src/Algorithm/conquer.jl
@@ -81,8 +81,7 @@ end
 """
     Coluna.Algorithm.RestrictedMasterHeuristic()
 
-Enforce integrality of master column variables and then solve the restricted master with 
-the optimizer attached to the master formulation.
+This algorithm enforces integrality of column variables in the master formulation and then solves the master formulation with its optimizer.
 """
 RestrictedMasterHeuristic() = 
     ParameterisedHeuristic(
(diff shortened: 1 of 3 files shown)


src/Algorithm/conquer.jl
```diff
@@ -80,8 +78,17 @@ struct ParameterisedHeuristic
     name::String
 end
 
-DefaultRestrictedMasterHeuristic() = 
-    ParameterisedHeuristic(RestrictedMasterIPHeuristic(), 1.0, 1.0, 1, 1000, "Restricted Master IP")
+"""
+    Coluna.Algorithm.RestrictedMasterHeuristic()
+
+Enforce integrality of master column variables and then solve the restricted master with 
+the optimizer attached to the master formulation.
```
laradicp, line 85: I got confused about specifying the restricted master, because I thought it meant master and restricted master were different here, and I didn't understand how. Without the "restricted" I probably wouldn't have made this confusion, but now I see that the way it's written is more accurate. So for me it would be easier to understand without any "restricted" or with all "master" specified, but I'm not sure if this change would be really better for a more experienced reader. It's clear for me now anyway.

laradicp (approved):

src/Algorithm/conquer.jl
guimarqu, line 85:
```suggestion
This algorithm enforces integrality of column variables in the master formulation and then solves the master formulation with its optimizer.
```

@laradicp  is it better ?

laradicp: Yes, it is

PR #547, opened 2021-06-18
Retrieve disaggregated solution

### LINKED ISSUE #540: Retrieve disagreggated solution
Currently, the user can retrieve only the solution of a model through the original variables. This is the projection of the master solution on the original formulation.

When a subproblem has an upper multiplicity greater than 1 and the optimal solution uses several solutions from this subproblem, all these subproblem solutions are aggregated. 
If the user has only access (as it is currently the case) to the solution of the original formulation, there is no straightforward way to get the number of subproblem solutions used and disaggregate the subproblem solutions.
However, if the user can retrieve the master solution, he will have all these information.

### Example : bin-packing

Let's consider this bin-packing problem. There is one type of bin with capacity $Q = 50$ and 20 items.
We must pack all the items using as few bins as possible while respecting the capacity of each bin.

We declare `y[k]` the number of bins of type `k` used.
We declare `x[k, i] = 1` if bin `i` is packed into a bin of type `k`.

```julia
using BlockDecomposition, Coluna, JuMP, GLPK

I = 1:20
@axis(BinsType, [1])

w = [10, 15, 27, 9, 12, 5, 17, 33, 4, 9, 34, 41, 26, 27, 16, 11, 19, 17, 19, 11]
Q = 50

coluna = JuMP.optimizer_with_attributes(
    Coluna.Optimizer,
    "params" => Coluna.Params(solver = Coluna.Algorithm.TreeSearchAlgorithm(
        branchingtreefile = "playgap.dot"
    )),
    "default_optimizer" => GLPK.Optimizer
)

model = BlockModel(coluna)

@variable(model, x[k in BinsType, i in I], Bin)
@variable(model, y[k in BinsType], Bin)

@constraint(model, sp[i in I], sum(x[k, i] for k in BinsType) == 1)
@constraint(model, ks[k in BinsType], sum(w[i] * x[k, i] for i in I) - y[k] * Q <= 0)

@objective(model, Min, sum(y[k] for k in BinsType))

@dantzig_wolfe_decomposition(model, dec, BinsType)
subproblems = BlockDecomposition.getsubproblems(dec)
specify!.(subproblems, lower_multiplicity = 0, upper_multiplicity = length(I)) # we use at most 20 bins

JuMP.optimize!(model)
```
When we retrieve the optimal solution, we can only do : 

```julia
julia> value.(y)
1-dimensional DenseAxisArray{Float64,1,...} with index sets:
    Dimension 1, BlockDecomposition.Axis{:BinsType, Int64}(:BinsType, BlockDecomposition.AxisId{:BinsType, Int64}[1])
And data, a 1-element Vector{Float64}:
 8.0

julia> value.(x)
2-dimensional DenseAxisArray{Float64,2,...} with index sets:
    Dimension 1, BlockDecomposition.Axis{:BinsType, Int64}(:BinsType, BlockDecomposition.AxisId{:BinsType, Int64}[1])
    Dimension 2, 1:20
And data, a 1×20 Matrix{Float64}:
 1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0
```

You note that `value.(x)` does not give us any information about the solution...

Interface we would like to implement to retrieve the disaggregated solution  :

```julia
for k in BinsType
    bins = BlockDecomposition.getsolutions(model, subproblem[k])
    for bin in bins
        lambda_val = BlockDecomposition.value(bin) # value of the master column variable
        for i in I
            x_val = BlockDecomposition.value(bin, x[k, i]) # coefficient of original var x[k, i] in the column knp
        end
    end
end
```

laradicp:
fix #540 

- `getsolutions` retrieves the disaggregated solution from `disagg_result` in `Optimizer`


guimarqu:
It's a good start.

The disaggregated solution may have columns with fractional values if we proved that there exists a combination that leads to feasible columns with integer values. Since we don't have such a mechanism, this solution may be infeasible.

In the meantime, we can check the integrality of columns from sps with upper multiplicity greater than 1 in the feasibility checker.

Let's keep this issue for later.

guimarqu:
By the way, I wouldn't store the `OptimizationState`s in a vector because it creates an implicit convention that we'll forget (1st entry for the aggregated solution, 2nd for the disaggregated one).

test/sol_disaggregation_tests.jl
```diff
@@ -0,0 +1,52 @@
+function sol_disaggregation_tests()
+    I = 1:20
```
guimarqu, line 2: I think that you should test the feature by creating a new algorithm that returns a pre-defined solution. Then, you test if you can retrieve the same solution with the new interface.

guimarqu (approved):

test/sol_disaggregation_tests.jl
guimarqu: (I misclicked)

guimarqu (approved):

PR #548, opened 2021-06-24
Unit tests for OptimizationState

laradicp:
Test in "test/unit/optimizationstate.jl" line 116 (`add_lp_dual_sol!`) was failing. For me it seemed that the sorting wasn't correct for dual solutions, that's why I reversed it.

### DIFF
--- a/test/unit/optimizationstate.jl
+++ b/test/unit/optimizationstate.jl
@@ -5,6 +5,9 @@ function optstate_unit_tests()
 end
 
 function update_sol_tests()
+    ############################################################################################
+    # MinSense                                                                                 #
+    ############################################################################################
     form = create_formulation!(Env(Coluna.Params()), Original())
     var = ClMP.setvar!(form, "var1", ClMP.OriginalVar)
     constr = ClMP.setconstr!(form, "constr1", ClMP.OriginalConstr)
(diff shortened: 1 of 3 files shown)


guimarqu:
Great ! What happens if the objective sense changes ?

laradicp:
> Great ! What happens if the objective sense changes ?

I'll add more tests to check

test/unit/optimizationstate.jl
```diff
@@ -0,0 +1,355 @@
...
+    # check that incumbent bound is updated
+    @test get_ip_primal_bound(state) == 2.0
+    update_ip_primal_sol!(state, PrimalSolution(
+        form, [getid(var)], [3.0], 3.0, CB.UNKNOWN_FEASIBILITY
+    ))
+    # check that solution worse than `primalsol` is NOT added to `state.ip_primal_sols`
+    for sol in get_ip_primal_sols(state), (varid, val) in sol
+        @test val == primalsol[varid]
+    end
+    ###
```
guimarqu, line 36: So if I understand well, in this test, you check that the only solution in `state.ip_primal_sols` is `primalsol`.

Do you think you can rewrite the test like the following one ? It's easier to understand, isn't it ?

```julia
@test length(get_ip_primal_sols(state)) == 1
@test get_ip_primal_sols(state)[1] == primalsol
```

laradicp: Yes, it is. But I think I would have to keep the second for loop anyway
```julia
for (varid, val) in get_ip_primal_sols(state)[1]
```
because when I tried to compare 2 primal solutions I got an error, I think it's due to the dictionary.

laradicp: Is it better this way?

guimarqu (approved):

PR #549, opened 2021-07-01
Solutions comparison operators

laradicp:
Add `Base.in(::Tuple{De,Va}, ::Solution{Mo,De,Va})` to allow `==` and `isequal` comparisons between solutions.

### DIFF
--- a/src/Coluna.jl
+++ b/src/Coluna.jl
@@ -3,7 +3,7 @@ module Coluna
 # Base functions for which we define more methods in Coluna
 import Base: isempty, hash, isequal, length, iterate, getindex, lastindex,
     getkey, delete!, setindex!, haskey, copy, promote_rule, convert, isinteger,
-    push!, filter, diff, hcat
+    push!, filter, diff, hcat, in
 
 import BlockDecomposition, MathOptInterface, TimerOutputs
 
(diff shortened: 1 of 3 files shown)


guimarqu:
Ok with the changes but what's the problem of defining `==` for `Solution`?

laradicp:
> Ok with the changes but what's the problem of defining `==` for `Solution`?

No problem. `Base.==` is not imported to Coluna yet, and I would implement like this
```julia
Base.==(s1::Solution, s2::Solution) = isequal(s1, s2)
```
So I didn't know if adding this extra method would be really preferable or if just `isequal` is as easily understandable.

guimarqu:
Documentation says that default implementation of `isequal` calls `==`.

These methods seems to have different behaviors when arguments are `Float64`.
In very special cases, we may have `isequal(sol1,sol2) !== (sol1 == sol2)`.

laradicp:
The problem was actually that `Base.in` is only defined for `Pair`. Since a `Tuple` is used for the `AbstractDict` in Coluna, it threw an error. I took the code from Base and adapted it to `Tuple`. Now both `==` and `isequal` work as expected.

guimarqu (approved):

PR #562, opened 2021-07-19
Follow up of "custom data": use BlockDecomposition methods `customvars!` and `customconstrs!`

laradicp:
task for #536

- Implement MOI set methods for custom data in Coluna
- Use `BlockDecomposition.AbstractCustomData`
- Add `custom_families_id` field to Env and pass the dict to FormulationManager


guimarqu (approved):
Great !

PR #566, opened 2021-07-21
Add support for `AffExpr, MathOptInterface.Interval{Float64}`

### LINKED ISSUE #563: Unsupported `AffExpr, MathOptInterface.Interval{Float64}`
Thanks for your answer.

When looking into it I noticed that the same error is thrown when I want to solve a model with a constraint of type `(AffExpr, MathOptInterface.Interval{Float64})` (e.g. because the model contains the constraint `1 <= x <= 2`).

_Originally posted by @davidmeichel in https://github.com/atoptima/Coluna.jl/issues/559#issuecomment-882078036_

laradicp:
fix #563 

I hoped to find a method like this in MOI:

```julia
function MOI.set(
    model::MOI.ModelLike,
    attr::MOI.AbstractConstraintAttribute,
    bridge::SplitIntervalBridge,
    value,
)
    MOI.set(model, attr, bridge.lower, value)
    MOI.set(model, attr, bridge.upper, value)
end
```

But I couldn't find any. The fallback we get (https://github.com/jump-dev/MathOptInterface.jl/blob/978096044b3ab89ae57e778c9369f2198dba49af/src/Bridges/bridge.jl#L120) seems to confirm it is necessary to add a new set method at least for `SplitIntervalBridge`.

### DIFF
--- a/src/MOIwrapper.jl
+++ b/src/MOIwrapper.jl
@@ -437,6 +437,15 @@ function MOI.set(
 end
 
 function MOI.set(
+    model::MOI.Bridges.LazyBridgeOptimizer{Coluna.Optimizer}, ::BD.ConstraintDecomposition,
+    bridge::MOI.Bridges.Constraint.SplitIntervalBridge, annotation::BD.Annotation
+)
+    MOI.set(model.model, BD.ConstraintDecomposition(), bridge.lower, annotation)
+    MOI.set(model.model, BD.ConstraintDecomposition(), bridge.upper, annotation)
+    return
+end
+
+function MOI.set(
     model::Coluna.Optimizer, ::MOI.ConstraintName, constrid::MOI.ConstraintIndex{F,S}, name::String
 ) where {F<:MOI.ScalarAffineFunction,S}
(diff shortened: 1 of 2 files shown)


guimarqu (approved):

laradicp:
If I remove this method here, I get this error now:
```julia
 MathOptInterface.UnsupportedAttribute{BlockDecomposition.ConstraintDecomposition}: Attribute BlockDecomposition.ConstraintDecomposition() is not supported by the model.
 ```
So maybe this is the bug?

**Comment by guimarqu** [comment]
Yes I'm currently fixing it. We need this method. The correct definition is specific to the ConstraintDecomposition attribute: 

```julia
function MOI.set(
    model::MOI.Bridges.LazyBridgeOptimizer{Coluna.Optimizer}, attr::BlockDecomposition.ConstraintDecomposition,
    bridge::MOI.Bridges.Constraint.SplitIntervalBridge, value
)
    MOI.set(model.model, attr, bridge.lower, value)
    MOI.set(model.model, attr, bridge.upper, value)
    return
end
```

and it should be in BlockDecomposition.

PR #568, opened 2021-07-26
Adding Node Finalizer

artalvpes:
Include a new custom algorithm, called node finalizer, that may conquer the node and/or find new feasible solutions. Some code let in the test function should still be encapsulated into additional API functions for a more user-friendly interface.


src/Algorithm/branching/branchingalgo.jl
```diff
@@ -127,7 +128,7 @@ function perform_strong_branching_with_phases!(
  
         for (group_index,group) in enumerate(groups)
             #TO DO: verify if time limit is reached
-            if phase_index == 1
```
guimarqu, line 130: what was the bug?

src/Algorithm/conquer.jl
```diff
@@ -91,6 +91,17 @@ ParamRestrictedMasterHeuristic() =
         1.0, 1.0, 1, 1000, "Restricted Master IP"
     )
 
+####################################################################
+#                      ParameterisedNodeFinalizer
+####################################################################
+
+struct ParameterisedNodeFinalizer
```
guimarqu, line 98: Can we call it `NodeFinalizer` or `ConquerPostProcessor` ?

src/Algorithm/conquer.jl
```diff
@@ -280,6 +291,62 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
                break
             end   
         end
+
+        # if the gap is still unclosed, try to run the node finalizer if any
+        if getterminationstatus(nodestate) != TIME_LIMIT &&
+            !ip_gap_closed(nodestate, atol = algo.opt_atol, rtol = algo.opt_rtol) &&
+            algo.node_finalizer !== nothing &&
+            getdepth(node) >= algo.node_finalizer.min_depth && 
+            mod(get_tree_order(node) - 1, algo.node_finalizer.frequency) == 0
```
guimarqu, line 300: Very difficult to read.

Consider something like 
```julia
run_node_finalizer = algo.node_finalizer !== nothing
run_node_finalizer &= getterminationstatus(nodestate) != TIME_LIMIT
run_node_finalizer &= !ip_gap_closed(nodestate, atol = algo.opt_atol, rtol = algo.opt_rtol)
run_node_finalizer &= getdepth(node) >= algo.node_finalizer.min_depth
run_node_finalizer &= mod(get_tree_order(node) - 1, algo.node_finalizer.frequency) == 0

if run_node_finalizer
    # ...
```



```

File: src/Algorithm/conquer.jl [hunk 4]
```diff
@@ -280,6 +291,62 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
+                recordids = store_records!(reform)
+            end   
+
+            nf_output = run!(nodefinalizer, env, reform, OptimizationInput(nodestate))
+            status = getterminationstatus(getoptstate(nf_output))
+            status == TIME_LIMIT && setterminationstatus!(nodestate, status)
+            ip_primal_sols = get_ip_primal_sols(getoptstate(nf_output))
+
+            # if the node has been conquered by the node finalizer
+            if (status in (OPTIMAL, INFEASIBLE))
```
guimarqu, line 317:
```suggestion
            if status in (OPTIMAL, INFEASIBLE)
```

src/Algorithm/conquer.jl
```diff
@@ -280,6 +291,62 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
+                # set the ip solutions found without checking the cuts and finish
+                best_primal_bound = PrimalBound(reform)
+                if ip_primal_sols !== nothing && length(ip_primal_sols) > 0
+                    for sol in sort(ip_primal_sols)
+                        update_ip_primal_sol!(nodestate, sol)
+                        primal_bound = PrimalBound(reform, getvalue(sol))
+                        if isbetter(primal_bound, best_primal_bound)
```
guimarqu, line 324: Is this condition always true since `best_primal_bound == PrimalBound(reform) == +/- Inf ` ?

src/Algorithm/conquer.jl
```diff
@@ -280,6 +291,62 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
+                            best_primal_bound = primal_bound
```
guimarqu, line 325: Where do you use `best_primal_bound` ?

src/Algorithm/conquer.jl
```diff
@@ -280,6 +291,62 @@ function run!(algo::ColCutGenConquer, env::Env, reform::Reformulation, input::Co
...
+                    # we start with worst solution to add all improving solutions
+                    for sol in sort(ip_primal_sols)
+                        cutgen = CutCallbacks(call_robust_facultative = false)
+                        # TO DO : Node finalizer should ensure itselves that the returned solution is feasible
+                        cutcb_output = run!(cutgen, env, getmaster(reform), CutCallbacksInput(sol))
+                        if cutcb_output.nb_cuts_added == 0
+                            update_ip_primal_sol!(nodestate, sol)
+                        end
+                    end
+                end
```
guimarqu, line 344: It seems like this section is never covered.

test/node_finalizer_tests.jl
```diff
@@ -0,0 +1,165 @@
+CL.@with_kw struct EnumerativeFinalizer <: ClA.AbstractOptimizationAlgorithm
```
guimarqu, line 1: Can you add a short description of the test?

src/Algorithm/branching/branchingalgo.jl
artalvpes: There is a "continue" above. Phase 1 may be skipped. If this occurs, "regenerate_children" in phase 2 will fail.

src/Algorithm/conquer.jl
artalvpes: Done.

src/Algorithm/conquer.jl
artalvpes: Done.

src/Algorithm/conquer.jl
artalvpes: No. There is a loop over the solutions returned by the node finalizer.

src/Algorithm/conquer.jl
artalvpes: You are right. This code was inherited from the primal heuristics. Cleaned it up.

src/Algorithm/conquer.jl
artalvpes: Julia does not implement short-circuit evaluation for "run_node_finalizer &= ...". It generated an error and I had to replace by "run_node_finalizer = run_node_finalizer && ...".

src/Algorithm/conquer.jl
artalvpes: It is because node finalizer was never returning a solution not proven optimal. Fixed.

src/Algorithm/conquer.jl
guimarqu: What was the bug? I could do : 

```julia
julia> a = true
true

julia> a &= true
true

julia> a &= false
false

julia> a &= a && a
false

julia> a &= a || a
false

julia> a &= !a
false
```

guimarqu: Did you try with parenthesis around each condition ?

guimarqu: Artur explained me, ok with the new version :)

guimarqu (approved):

PR #569, opened 2021-08-04
Update callbacks documentation

guimarqu:


docs/src/man/callbacks.jl
```diff
@@ -0,0 +1,140 @@
+# # Callbacks
+
+# Callbacks are functions defined by the user that allow him to take over part of the default 
+# algorithm. 
+# The more classical callbacks in a branch-and-price solver are:
+
+# - Pricing callback that takes over the procedure to determine whether the current master LP 
+#     solution is optimum or produces an entering variable with negative reduced cost
```
laradicp, line 8: I understand this, but at first I was a little confused because subproblems weren't mentioned. Then it was clearer when I read lines 19-22.

laradicp (approved):

docs/src/man/callbacks.jl
guimarqu:
```suggestion
#     solution is optimum or produces an entering variable with negative reduced cost by solving subproblems
```

PR #571, opened 2021-08-06
MOI tests: LP unit tests and contlinear tests

laradicp:


src/MOIwrapper.jl
```diff
@@ -302,6 +301,8 @@ function MOI.get(
     orig_form = get_original_formulation(model.inner)
     constrid = getid(model.constrs[index])
     terms = MOI.ScalarAffineTerm{Float64}[]
+    coefmatrix = getcoefmatrix(orig_form)
+    coefmatrix.fillmode && closefillmode!(coefmatrix)
```
guimarqu, line 305: For what test did you have to do that (is it a test that does set get set) ? 

If you close the fill mode, then writing operations become time consuming. Do you think we should we be able to reopen the fill mode ?

laradicp: Tests "linear1" and "linear6" threw this error:
```
Expression: MOI.get(model, MOI.ConstraintFunction(), c1) ≈ fx
View of a row not available in fill mode (Open an issue at https://github.com/atoptima/DynamicSparseArrays.jl if you need it).
```
I don't know if there's a better fix to that and I don't understand fully how fill mode works yet, but being able to reopen the fill mode seems like a good option.

guimarqu (approved):

Issue #583, opened 2021-08-10
Constraint on variable overwritten


laradicp:
**Describe the bug**
When adding a new constraint on variable, it may overwrite previous constraints of the same type.

**To Reproduce**
```julia
# x >= 1
c1 = MOI.add_constraint(
    model,
    MOI.SingleVariable(x),
    MOI.GreaterThan(1),
)
# x >= 2
c2 = MOI.add_constraint(
    model,
    MOI.SingleVariable(x),
    MOI.GreaterThan(2),
)
```
Impossible to delete `c2` without losing `c1` because `Coluna._constraint_on_variable!` called by `MOI.add_constraint` changes the perennial bounds of the variables.

**Additional context**
Maybe related to the incomplete dual solution (missing `MOI.ConstraintDual` for `SingleVariable` constraints) and to https://github.com/atoptima/Coluna.jl/issues/531.

guimarqu:
Coluna will now an error `MOI.LowerBoundAlreadySet`. (see #628)

Issue #531, opened 2021-05-20
Variable bounds not updated in branching


laradicp:
**Describe the bug**
Branching algorithm adds constraints for variable bounds (in function `generate_children`), but the bounds of the variable itself are not updated. The variable bounds retrieved from `BlockDecomposition.callback_lb(cbdata, var)` and `BlockDecomposition.callback_ub(cbdata, var)` are incorrect.

**To Reproduce**
1. Optimize a model with `maxnumnodes = 2`
2. Check the branching constraint (variable and rhs)
3. ```
    cbdata = MathProg.PricingCallbackData(model.moi_backend.inner.re_formulation.master)
    @test BD.callback_lb(cbdata, variable) == rhs
    @test BD.callback_ub(cbdata, variable) == rhs
    ```

**Expected behavior**
If the branching constraint is greater or equal, first test will throw an error. Second otherwise. The bound retrieved is always a perene bound.

**Suggestion**
`setcurlb!` and `setcurub!` as soon as the variables of the child nodes are accessible.

**Question**
Are there specific variables to each node? If not, maybe it would be possible to retrieve the bounds from `Node.branchdescription` instead of from `BlockDecomposition.callback_lb(cbdata, var)` and `BlockDecomposition.callback_ub(cbdata, var)`.

Issue #436, opened 2021-01-27
MOI tests


guimarqu:
https://jump.dev/MathOptInterface.jl/v0.9.14/apimanual/#Testing-guideline-1

TODO : 

- [x] Basic unit tests (done in #437)
- [x] MIP unit tests
- [x] `intlinear` tests (with `SolveIpForm` as top-solver)
- [ ] LP unit tests
- [ ] `contlinear` tests (with `SolveLpForm` as top-solver)

guimarqu:
close in favor of #620

PR #590, opened 2021-08-14
Better support of single variable constraints

guimarqu:


guimarqu:
I think it's time to merge this PR into master... I think the review will be a very hard job.

This PR does not fix any issue but is a prerequisite for issues #583 #593 #436 #591 #531.

Moreover, it turns out that `SingleVariableConstraint` disappears in the next release of MathOptInterface. 
Without this PR, this change in MOI would dramatically decrease the performance of Coluna. I'll update the MOI wrapper in a future PR.

src/Algorithm/node.jl
```diff
@@ -8,7 +8,7 @@ mutable struct Node
     depth::Int
     parent::Union{Nothing, Node}
     optstate::OptimizationState
-    #branch::Union{Nothing, Branch} # branch::Id{Constraint}
+    #branch::Union{Nothing, Branch} # branch::ConstrId
```
rrsadykov, line 11: This can be deleted

rrsadykov (approved):
Yes, I think that it is not possible to really review this pull request. You may just explain us (at a high level) what did you do on our next meeting.

guimarqu:
@rrsadykov Yes it's planned ! I don't merge it right now because I think it's better to release a 0.3.10 before. More work for 0.4.0 than I initially thought.

PR #594, opened 2021-09-04
Give access to the dual solution of the master to the user

### LINKED ISSUE #591: Access to dual value of constraints of the master problem
Hello,

I would like to access the duals associated with the constraints in the master problem for a separate model, and it seems like I can't if they are added as lazy constraints. (I could be completely wrong about this, but I did not see a solution when looking into the topic.) 

Would it be possible to iteratively add constraints to the master problem without the use of lazy constraints, perhaps something similar to @constraint() in JuMP? 

Thank you so much!

guimarqu:
fix #591

### DIFF
--- a/src/Algorithm/colgen.jl
+++ b/src/Algorithm/colgen.jl
@@ -57,7 +57,6 @@ Here are their meanings :
 - `DB` is the dual bound of the master LP at the current iteration
 - `mlp` is the objective value of the master LP at the current iteration
 - `PB` is the objective value of the best primal solution found by Coluna at the current iteration
-
 """
 @with_kw struct ColumnGeneration <: AbstractOptimizationAlgorithm
     restr_master_solve_alg = SolveLpForm(get_dual_solution=true)
@@ -691,6 +690,11 @@ function cg_main_loop!(
             lp_bound = get_lp_primal_bound(rm_optstate) + getvalue(partial_solution)
             set_lp_primal_bound!(cg_optstate, lp_bound)
 
+            dual_rm_sol = get_best_lp_dual_sol(rm_optstate)
+            if dual_rm_sol !== nothing
+                set_lp_dual_sol!(cg_optstate, dual_rm_sol)
+            end
+
             if phase != 1 && !contains(rm_sol, varid -> isanArtificialDuty(getduty(varid)))
                 proj_sol = proj_cols_on_rep(rm_sol, masterform)
                 if isinteger(proj_sol) && isbetter(lp_bound, get_ip_primal_bound(cg_optstate))
@@ -774,8 +778,6 @@ function cg_main_loop!(
         update_stab_after_colgen_iteration!(stabunit)
 
         dual_bound = get_ip_dual_bound(cg_optstate)
-        primal_bound = get_lp_primal_bound(cg_optstate)
-        ip_primal_bound = get_ip_primal_bound(cg_optstate)
 
         if ip_gap_closed(cg_optstate, atol=algo.opt_atol, rtol=algo.opt_rtol)
             setterminationstatus!(cg_optstate, OPTIMAL)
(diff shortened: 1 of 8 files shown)


src/Algorithm/colgen.jl
```diff
@@ -691,6 +690,11 @@ function cg_main_loop!(
             lp_bound = get_lp_primal_bound(rm_optstate) + getvalue(partial_solution)
             set_lp_primal_bound!(cg_optstate, lp_bound)
 
+            dual_rm_sol = get_best_lp_dual_sol(rm_optstate)
+            if dual_rm_sol !== nothing
+                set_lp_dual_sol!(cg_optstate, dual_rm_sol)
```
rrsadykov, line 695: Is it really necessary to set the primal and dual solution to `cg_optstate` on every column generation iteration? May be ne may do it only at the end of column generation to save time?

src/Algorithm/treesearch.jl
```diff
@@ -88,6 +88,7 @@ mutable struct TreeSearchRuntimeData{Sense}
     Sense::Type{<:Coluna.AbstractSense}
     conquer_units_to_restore::UnitsUsage
     worst_db_of_pruned_node::DualBound{Sense}
+    worst_dual_sol_of_pruned_node::Union{Nothing,DualSolution}
```
rrsadykov, line 91: I do not understand the purpose of `worst_db_of_pruned_node` and `worst_dual_sol_of_pruned_node`. Why do we need that? Normally, the current dual bound of TreeSearch is computed as the worst dual bound of open nodes.

rrsadykov (changes requested):
For me, the top algorithm should return only one OptimizationState which contains the best found IP solution and possibly the LP primal and dual solutions for the root (this may be parameterized to save time).

src/Algorithm/treesearch.jl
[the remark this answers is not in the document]
guimarqu: I don't know why I did that.

src/optimize.jl
```diff
@@ -110,6 +110,13 @@ function optimize!(
         add_lp_primal_sol!(outstate, proj_cols_on_rep(lp_primal_sol, master))
     end
 
+    # lp_dual_sol to retrieve, for instance, the dual value of generated cuts
+    # TODO: how to project the master dual solution on the original formulation ?
```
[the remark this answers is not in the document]
guimarqu: Ok for me.

src/optimize.jl
```diff
@@ -110,6 +110,13 @@ function optimize!(
...
+    lp_dual_sol = get_best_lp_dual_sol(algstate)
```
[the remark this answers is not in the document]
guimarqu: Ok, I had doubts about what I did and what you suggest, so let's return the dual solution found at the root node.

src/Algorithm/colgen.jl
guimarqu: I don't think it will change anything for performances. It's just a pointer that you put in the optimisation state.

src/Algorithm/treesearch.jl
```diff
@@ -308,31 +309,25 @@ function updatedualbound!(data::TreeSearchRuntimeData)
...
         db = get_ip_dual_bound(getoptstate(node))
-        sol = get_best_lp_dual_sol(getoptstate(sol))
         if isbetter(worst_bound, db)
             worst_bound = db
-            worst_sol = sol
         end
     end
+
     if isbetter(worst_bound, data.worst_db_of_pruned_node)
         worst_bound = data.worst_db_of_pruned_node
```
rrsadykov, line 327: I still do not understand why dual bound of the tree search takes into account dual bounds of pruned nodes? It should be just equal to the worst dual bound of open nodes (or to the primal bound if there are no open nodes)

rrsadykov (changes requested):

src/Algorithm/treesearch.jl
guimarqu: This is something that we should address in a separate PR. I neither don't understand why I did that.

rrsadykov: Why separate PR? This is critical issue. It should be addressed before the next release.

guimarqu: Because this is not the scope of this PR. There is no rush for the next release.

rrsadykov (approved):

PR #598, opened 2021-09-08
dual bound of tree search is dual bound of open nodes

### LINKED ISSUE #597: Dual bound of tree search should be equal to the worst dual bound of open nodes
I still do not understand why dual bound of the tree search takes into account dual bounds of pruned nodes? It should be just equal to the worst dual bound of open nodes (or to the primal bound if there are no open nodes)

_Originally posted by @rrsadykov in https://github.com/atoptima/Coluna.jl/pull/594#discussion_r704136406_

guimarqu:
fix #597

### DIFF
--- a/test/full_instances_tests.jl
+++ b/test/full_instances_tests.jl
@@ -98,6 +98,31 @@ function generalized_assignment_tests()
         @test CLD.GeneralizedAssignment.print_and_check_sol(data, model, x)
     end
 
+
+    @testset "gap - node limit" begin
+        data = CLD.GeneralizedAssignment.data("mediumgapcuts3.txt")
+
+        coluna = JuMP.optimizer_with_attributes(
+            CL.Optimizer,
+            "params" => CL.Params(
+                solver = ClA.TreeSearchAlgorithm(
+                    maxnumnodes = 5
+                )
+            ),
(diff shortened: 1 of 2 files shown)


test/full_instances_tests.jl
```diff
@@ -98,6 +98,31 @@ function generalized_assignment_tests()
...
+            "default_optimizer" => GLPK.Optimizer
+        )
+
+        model, x, dec = CLD.GeneralizedAssignment.model(data, coluna)
+        BD.objectiveprimalbound!(model, 2000.0)
+        BD.objectivedualbound!(model, 0.0)
+
+        JuMP.optimize!(model)
+
+        @test JuMP.objective_bound(model) ≈ 1547.3889
```
rrsadykov, line 121: Are you sure that this test is reproducible? Branching decisions may be different depending on the fractional solution (which depends on the LP solver).

rrsadykov (approved):

test/full_instances_tests.jl
guimarqu: Yes, I'm not satisfied with this test, I'm leaving it until we think about better tests.

guimarqu: If it fails for some reason, we can replace `@test`  by `@test_broken`

Issue #478, opened 2021-03-25
Shortcut for `TreeSearchAlgorithm` configuration


guimarqu:
```julia
bcp = BranchCutAndPrice(...)
```

will call `TreeSearchAlgorithm`

PR #603, opened 2021-09-13
Alias for a simplified BCP parameterisation 

rrsadykov:
Introduces `Coluna.Algorithm.BranchCutAndPriceAlgorithm` alias for a simplified parameterisation of the branch-cut-and-price algorithm.

Addresses issue #478.

### DIFF
--- a/src/Algorithm/branchcutprice.jl
+++ b/src/Algorithm/branchcutprice.jl
@@ -11,10 +11,9 @@
         colgen_stabilization::Float64 = 0.0, 
         colgen_cleanup_threshold::Int = 10000,
         colgen_stages_pricing_solvers::Vector{Int} = [1],
-        stbranch_phases_num_candidates::Int = [],
-        stbranch_intrmphase_colgen_stages::Vector{Int} = [1],
-        stbranch_intrmphase_pricing_solvers::Vector{Int} = [1],
-        stbranch_intrmphase_colgen_maxiters::Vector{Int} = [100])
+        stbranch_phases_num_candidates::Vector{Int} = Vector{Int}(),
+        stbranch_intrmphase_stages::Vector{NamedTuple{(:userstage, :solverid, :maxiters), Tuple{Int64, Int64, Int64}}}
+    )
 
 Coluna.Algorithm.BranchCutAndPriceAlgorithm is an alias for a simplified parameterisation 
 of the branch-cut-and-price algorithm.
--- a/src/Algorithm/branchcutprice.jl
+++ b/src/Algorithm/branchcutprice.jl
@@ -1,0 +1,192 @@
+"""
+
+    Coluna.Algorithm.BranchCutAndPriceAlgorithm(;
+        maxnumnodes::Int = 100000,
+        opt_atol::Float64 = Coluna.DEF_OPTIMALITY_ATOL,
+        opt_rtol::Float64 = Coluna.DEF_OPTIMALITY_RTOL,
+        restmastipheur_timelimit::Int = 600,
+        restmastipheur_frequency::Int = 1,
+        restmastipheur_maxdepth::Int = 1000,
+        max_nb_cut_rounds::Int = 3,
+        colgen_stabilization::Float64 = 0.0, 
+        colgen_cleanup_threshold::Int = 10000,
+        colgen_stages_pricing_solvers::Vector{Int} = [1],
+        branching_priorities::Vector{Float64} = [1.0],
+        stbranch_phases_num_candidates::Int = [],
+        stbranch_intrmphase_colgen_stages::Vector{Int} = [1],
+        stbranch_intrmphase_pricing_solvers::Vector{Int} = [1],
+        stbranch_intrmphase_colgen_maxiters::Vector{Int} = [100])
+
+Coluna.Algorithm.BranchCutAndPriceAlgorithm is an alias for a simplified parameterisation 
+of the branch-cut-and-price algorithm.
+
+Parameters : 
+- `maxnumnodes` : maximum number of nodes explored by the algorithm
+- `opt_atol` : optimality absolute tolerance 
+- `opt_rtol` : optimality relative tolerance
+- `restmastipheur_timelimit` : time limit in seconds for the restricted master heuristic 
+                               (if <= 0 then the heuristic is disabled)
+- `restmastipheur_frequency` : frequency of calls to the restricted master heuristic                               
+- `restmastipheur_maxdepth` : maximum depth of the search tree when the restricted master heuristic is called
+- `max_nb_cut_rounds` : maximum number of cut generation rounds in every node of the search tree
+- `colgen_stabilization` : parameterisation of the dual price smoothing stabilization of column generation
+                           0.0 - disabled, 1.0 - automatic, ∈(0.0,1.0) - fixed smoothing parameter
+- `colgen_cleanup_threshold` : threshold (number of active columns) to trigger the restricted master LP clean up
+- `colgen_stages_pricing_solvers` : vector of pricing solver ids for every column generation stage,
+                                    pricing solvers should be specified using argument `solver` of `BlockDecomposition.specify!()`,
+                                    the number of column generation stages is equal to the length of this vector,
+                                    column generation stages are executed in the reverse order,
+                                    the first stage should be exact to ensure the optimality of the BCP algorithm
+- `branching_priorities` : vector of different branching priorities (in decreasing order) set for the variables of the model,
+                           branching priorities for variables can be set using `BlockDecomposition.branchingpriority!()`,
+                           branching candidates are generated in the decreasing order of branching priority
+                           if no branching candidates are found for a larger branching priority, the next lower branching priority is considered,
+                           in strong branching, branching candidates of the next lower priority are considered if
+                           - no branching candidates of a larger priority are generated
+                           - the maximum number of candidates for the first phase of strong branching is not yet reached
+                             and the next lower priority is not less than the rounded down value of an already generated branching candidate
+- `stbranch_phases_num_candidates` : maximum number of candidates for each strong branching phase, 
+                                     strong branching is activated if this vector is not empty,
+                                     the number of phases in strong branching is equal to min{3, length(stbranch_phases_num_candidates)},
+                                     in the last phase, the standard column-and-cut generation procedure is run,
+                                     in the first phase (if their number is >= 2), only the restricted master LP is resolved, 
+                                     in the second (intermediate) phase (if their number is >= 3), usually a heuristic pricing is used
+                                     or the number of column generation iterations is limited, this is parameterised with the three
+                                     next parameters, cut separation is not called in the intermediate strong branching phase, 
+                                     if the lenth of this vector > 3, then all values except first, second, and last ones are ignored
+- `stbranch_intrmphase_colgen_stages` : the number of column generation stages in the intemediate phase of strong branching 
+                                        is equal to the length of this vector, the values of this vector determine the `stage`
+                                        parameter passed to the pricing callback on every stage,
+                                        as before column generation stages are executed in the reverse order
+- `stbranch_intrmphase_pricing_solvers` : sets the solver id for every column generation stage in the intermediate phase of strong branching,
+                                          the length of this vector should be equal to length(stbranch_intrmphase_colgen_stages)
+- `stbranch_intrmphase_colgen_maxiters` : sets the maximum number of column generation iterations for every stage 
+                                          in the intemediate phase of strong branching,
+                                          the length of this vector should be equal to length(stbranch_intrmphase_colgen_stages)
+"""
+
+
+
+function BranchCutAndPriceAlgorithm(;
+        maxnumnodes::Int = 100000,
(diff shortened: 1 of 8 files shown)


rrsadykov:
Codecov Report is wrong

guimarqu:
I think that before merging this new interface, we must decide on how we define the branching priorities.

My suggestion is to define branching priorities only through BlockDecomposition. Currently, we only have `BlockDecomposition.branchingpriority!` to define the branching priority on a JuMP variable. I think it can replace `VarBranchingRule`. However, if the latter disappears, `AbstractBranchingRule` and `PrioritisedBranchingRule` also disappear.

The aim of `AbstractBranchingRule` is not clear to me.

rrsadykov:
`AbstractBranchingRule` is the algorithm type. Algorithms of this type find branching candidates. `VarBranchingRule` is a particular algorithm to perform branching. There can be other branching rules: Ryan&Foster, branching on constraints (defined by ``@expression` in JuMP), or the user can define his own branching rules. May be we should rename `BranchingRule` to `BranchingAlgorithm` to be more clear. 

The main problem here is that we mix variable branching priorities given using `BlockDecomposition` and branching algorithms defined in Coluna (these two parameterisations should be "compatible"). We cannot have just variable branching priorities, as it is important to be able to do other kinds of branching (non-robust branching is sometimes necessary for completeness).   

One solution may be to define variable branching using names, like 

```
branchingFirstOnXthenOnY = ClA.StrongBranching(
            phases = [] ,
            rules = [ClA.PrioritisedBranchingRule(ClA.VarBranchingRule(varname = "X"),  2.0, 2.0),
                         ClA.PrioritisedBranchingRule(ClA.VarBranchingRule(varname = "Y"), 1.0, 1.0)]
        )
```

For this, we need to be able to get the JuMP name of a variable (or expression) inside the branching algorithm in Coluna.

Then, branching priorities given using `BlockDecomposition` will serve just to sort individual variables with the same name. So, the "upper level" branching priorities will be given in the Coluna parameterisation, and "lower level" branching priorities will be given using `BlockDecomposition`. Actually, this is completely compatible with the current implementation. 

One may also parameterise `VarBranchingRule` with the selection strategy (most violated, smallest index, etc). This is how it is done in constraint programming solvers.

guimarqu:
Ok, thanks for the explanation. I had doubts about the aim of `AbstractBranchingRule` because there is currently only one rule implemented & because the selection criterion is in `StrongBranching`.

About your suggestion, I don't think we can rely on MathOptInterface to always provide us the name of the variables and constraints. Currently, if you build your model using the caching optimizer (which is the default mode), Coluna does not receive the name of variables and constraints from MathOptInterface. 
Moreover, users may want to use [bridges](https://jump.dev/MathOptInterface.jl/v0.10/submodules/Bridges/overview/) to transform variables and constraints unsupported by Coluna into supported ones.
I guess that some bridges may create new variables but how do you propagate the branching rules to these new variables with your suggestion ? I think that these new variables have no name.

rrsadykov:
Receiving names of variables (and constraints) is critical. In many cases, the user needs to know on which variable the branching is performed. 

In the worst case the name should be passed through `BlockDecomposition`. One may add variable name parameter in `BlockDecomposition.branchingpriority!()` or add another function `BlockDecomposition.varname!()`.

rrsadykov:
> Moreover, users may want to use [bridges](https://jump.dev/MathOptInterface.jl/v0.10/submodules/Bridges/overview/) to transform variables and constraints unsupported by Coluna into supported ones.
> I guess that some bridges may create new variables but how do you propagate the branching rules to these new variables with your suggestion ? I think that these new variables have no name.

This is too advanced for me. Created variables may just receive some default (f.e. empty) name. The user then knows that branching is done on a variables not defined by him.

guimarqu:
I agree that showing names of variables in branching is very useful when you implement a new application or in the R&D phase. However, I don't think we need them in production, so it's not so critical nor a requirement.
 
I guess created variables receive an empty name. However, I think we should be able to propagate the branching rule defined on the original variables because the latter may disappear. This propagation is done by MathOptInterface with a bridge that we will define for the BlockDecomposition.BranchingPriority attribute.
If we define branching rules in the algorithm, we can't do the propagation anymore.

I think it's very important to not try to bypass what has been done in MathOptInterface because we'll gain nothing except new bugs. The name should be provided by JuMP and MathOptInterface, not by BlockDecomposition.

rrsadykov:
> However, I don't think we need them in production, so it's not so critical nor a requirement.

In research, there is not production phase. 

> The names should be provided by JuMP and MathOptInterface, not by BlockDecomposition

I completely agree, but you said that we cannot count on that. 

I general, I do not understand what do you suggest to do with branching.

guimarqu:
I think we can define both branching priorities and branching rules through BlockDecomposition

```julia
branchingpriority!(x; rule = Coluna.VarBranchingRule(), root_priority = 1, non_root_priority = 2)
```

Before starting the optimization, Coluna can group variables that have the same branching rules and names are not a requirement anymore.

src/Algorithm/branchcutprice.jl
```diff
@@ -0,0 +1,166 @@
...
+        opt_atol::Float64 = Coluna.DEF_OPTIMALITY_ATOL,
+        opt_rtol::Float64 = Coluna.DEF_OPTIMALITY_RTOL,
+        restmastipheur_timelimit::Int = 600,
+        restmastipheur_frequency::Int = 1,
+        restmastipheur_maxdepth::Int = 1000,
+        max_nb_cut_rounds::Int = 3,
+        colgen_stabilization::Float64 = 0.0, 
+        colgen_cleanup_threshold::Int = 10000,
+        colgen_stages_pricing_solvers::Vector{Int} = [1],
+        stbranch_phases_num_candidates::Vector{Int} = Vector{Int}(),
```
guimarqu, line 14:
```suggestion
        stbranch_phases_num_candidates::Vector{Int} = Int[],
```

src/Algorithm/branchcutprice.jl
```diff
@@ -0,0 +1,166 @@
...
+        stbranch_intrmphase_stages::Vector{NamedTuple{(:userstage, :solverid, :maxiters), Tuple{Int64, Int64, Int64}}}
+    )
+
+Coluna.Algorithm.BranchCutAndPriceAlgorithm is an alias for a simplified parameterisation
```
guimarqu, line 18:
```suggestion
Alias for a simplified parameterisation 
```

src/Algorithm/branchcutprice.jl
```diff
@@ -0,0 +1,166 @@
...
+        opt_atol::Float64 = Coluna.DEF_OPTIMALITY_ATOL,
+        opt_rtol::Float64 = Coluna.DEF_OPTIMALITY_RTOL,
+        restmastipheur_timelimit::Int = 600,
+        restmastipheur_frequency::Int = 1,
+        restmastipheur_maxdepth::Int = 1000,
+        max_nb_cut_rounds::Int = 3,
+        colgen_stabilization::Float64 = 0.0, 
+        colgen_cleanup_threshold::Int = 10000,
+        colgen_stages_pricing_solvers::Vector{Int64} = [1],
+        stbranch_phases_num_candidates::Vector{Int64} = Vector{Int64}(),
```
guimarqu, line 67:
```suggestion
        stbranch_phases_num_candidates::Vector{Int64} = Int[],
```

src/Algorithm/branchcutprice.jl
```diff
@@ -0,0 +1,166 @@
...
+        stbranch_intrmphase_stages::Vector{NamedTuple{(:userstage, :solverid, :maxiters), Tuple{Int64, Int64, Int64}}} = [(userstage=1, solverid=1, maxiters=100)]
+)
+    heuristics::Vector{ParameterisedHeuristic} = []
```
guimarqu, line 70:
```suggestion
    heuristics = ParameterisedHeuristic[]
```

src/Algorithm/branchcutprice.jl
```diff
@@ -0,0 +1,166 @@
...
+                get_dual_bound = false,
+                time_limit = restmastipheur_timelimit
+            )),
+            1.0, 1.0, restmastipheur_frequency, 
+            restmastipheur_maxdepth, "Restricted Master IP"
+        )
+        push!(heuristics, heuristic)
+    end
+
+    colgen_stages::Vector{ColumnGeneration} = []
```
guimarqu, line 83:
```suggestion
    colgen_stages = ColumnGeneration[]
```

src/Algorithm/branchcutprice.jl
```diff
@@ -0,0 +1,166 @@
...
+    conquer = ColCutGenConquer(
+        stages = colgen_stages,
+        max_nb_cut_rounds = max_nb_cut_rounds,
+        primal_heuristics = heuristics,
+        opt_atol = opt_atol,
+        opt_rtol = opt_rtol
+    )
+
+    branching = NoBranching()
+    branching_rules::Vector{PrioritisedBranchingRule} = [PrioritisedBranchingRule(VarBranchingRule(), 1.0, 1.0)]
```
guimarqu, line 112:
```suggestion
    branching_rules = PrioritisedBranchingRule[PrioritisedBranchingRule(VarBranchingRule(), 1.0, 1.0)]
```

src/Algorithm/branchcutprice.jl
```diff
@@ -0,0 +1,166 @@
...
+
+    if !isempty(stbranch_phases_num_candidates)
+        branching_phases::Vector{BranchingPhase} = []
```
guimarqu, line 115:
```suggestion
        branching_phases = BranchingPhase[]
```

src/Algorithm/branchcutprice.jl
```diff
@@ -0,0 +1,166 @@
...
+        if length(stbranch_phases_num_candidates) >= 2
+            push!(branching_phases, 
+                BranchingPhase(first(stbranch_phases_num_candidates), RestrMasterLPConquer())
+            )    
+            if length(stbranch_phases_num_candidates) >= 3
+                intrmphase_stages::Vector{ColumnGeneration} = []
```
guimarqu, line 121:
```suggestion
                intrmphase_stages = ColumnGeneration[]
```

src/Algorithm/branching/varbranching.jl
```diff
@@ -85,18 +85,30 @@ function run!(
     master = getmaster(reform)
     groups = Vector{BranchingGroup}()
     local_id = input.local_id
+    selected_vars = Vector{Pair{VarId, Float64}}()
```
guimarqu, line 88:
```suggestion
    selected_vars = Pair{VarId, Float64}[]
```

src/Algorithm/branching/varbranching.jl
```diff
@@ -85,18 +85,30 @@ function run!(
...
+    max_priority = -Inf
     for (var_id, val) in input.solution
         # Do not consider continuous variables as branching candidates
         getperenkind(master, var_id) == Continuous && continue
-        getbranchingpriority(master, var_id) < input.minimum_priority && continue
+        #getbranchingpriority(master, var_id) < input.minimum_priority && continue
```
guimarqu, line 93:
```suggestion
```

src/Algorithm/branching/varbranching.jl
```diff
@@ -85,18 +85,30 @@ function run!(
...
         if !isinteger(val, input.int_tol)
-            #description string is just the variable name
-            candidate = VarBranchingCandidate(getname(master, var_id), var_id)
-            local_id += 1
-            push!(groups, BranchingGroup(candidate, local_id, val))
+            brpriority = getbranchingpriority(master, var_id)
+            if (brpriority > max_priority)
```
guimarqu, line 96:
```suggestion
            if brpriority > max_priority
```

src/Algorithm/branching/varbranching.jl
```diff
@@ -85,18 +85,30 @@ function run!(
...
+                max_priority = brpriority
+                selected_vars = [(var_id, val)]
+            elseif (brpriority == max_priority)
```
guimarqu, line 99:
```suggestion
            elseif brpriority == max_priority
```

src/Algorithm/branching/varbranching.jl
```diff
@@ -85,18 +85,30 @@ function run!(
...
+        #getbranchingpriority(master, var_id) < input.minimum_priority && continue
         if !isinteger(val, input.int_tol)
-            #description string is just the variable name
-            candidate = VarBranchingCandidate(getname(master, var_id), var_id)
-            local_id += 1
-            push!(groups, BranchingGroup(candidate, local_id, val))
+            brpriority = getbranchingpriority(master, var_id)
+            if (brpriority > max_priority)
+                max_priority = brpriority
+                selected_vars = [(var_id, val)]
```
guimarqu, line 98: Does it mean that you create a new array each time you find a non-integer variable that has a greater priority than the current greatest priority?

I personally think it's better to loop twice over the solution. A first time to get the greatest branching priority, a second time to fill the `groups` array.

rrsadykov: Ok, I have modified the code in the latest commit as you suggested

guimarqu (approved):

PR #606, opened 2021-09-28
Docstrings in branching algorithms

guimarqu:
- selection criterion is a type
- removing methods and fields not used
- adding docstrings

### DIFF
--- a/src/Algorithm/branching/branchinggroup.jl
+++ b/src/Algorithm/branching/branchinggroup.jl
@@ -1,31 +1,46 @@
+############################################################################################
+# Branching Candidates
+############################################################################################
+
 """
     AbstractBranchingCandidate
 
-    A branching candidate should contain all information needed to generate node's children    
-    Branching candiates are also used to store the branching history. 
-    History of a branching candidate is a collection of statistic records for every time this branching
-        candidate was used to generate children nodes 
-    Every branching candidate should contain a description, i.e. a string which serves for printing purposed,
-    and also to detect the same branching candidates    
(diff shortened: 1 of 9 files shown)


src/Algorithm/branching/branchinggroup.jl
```diff
@@ -1,31 +1,46 @@
...
-    Contains a branching candidate together with additional "local" information needed during current branching
+A branching group is the union of a branching candidate and additional information that are
+computed during the execution of the branching algorithm (TODO : which one ?).
 """
 mutable struct BranchingGroup
-    candidate::AbstractBranchingCandidate
+    candidate::AbstractBranchingCandidate # the left-hand side in general.
     local_id::Int64
     lhs::Float64
-    fromhistory::Bool
```
rrsadykov, line 28: This field will be needed when the branching history is implemented (may be just to comment it now?)

src/Algorithm/branching/varbranching.jl
```diff
@@ -1,17 +1,24 @@
...
+    SingleVarBranchingCandidate
 
-    Contains a variable on which we branch
+It is an implementation of AbstractBranchingCandidate.
+This is the type of branching candidates produced by the branching rule 
+`SingleVarBranchingRule`.
 """
-struct VarBranchingCandidate <: AbstractBranchingCandidate
-    description::String
+struct SingleVarBranchingCandidate <: AbstractBranchingCandidate
```
rrsadykov, line 12: I think `VarBranchingCandidate` already indicates that we branch on a single variable, no?

PR #677, opened 2022-08-06
Support Representatives

guimarqu:
@artalvpes it will fix the error you sent. Implementing these two methods is the first step to support representatives in Coluna.


guimarqu:
Waiting for release of BlockDecomposition 1.10 https://github.com/atoptima/BlockDecomposition.jl/commit/af93067821ccdde13a66f2d4aa5a5ccd59184f1a

PR #688, opened 2022-08-16
Simplified version of storage

guimarqu:
- Previous storage was embedding a "manual" garbage collector. I simplified the code to remove the manual GC because Julia GC bound memory performances.
- UnitPermissions are still present but not used anymore. It did not create errors in e2e tests.
- Interface to implement unit storages and records is quite the same


guimarqu:
Broken tests because of JuMP 1.2.0 release.

PR #712, opened 2022-09-06
Simplify linear algebra operations syntax in colgen

guimarqu:
- data structure to store pre-computed information for the subgradient
- replacing all custom code for linear algebra with a generic one
- restructuring a little bit the `MP/solutions.jl` file (methods on abstract solution/primal solution/dual solution/linear algebra)


src/Algorithm/colgen.jl
```diff
@@ -288,6 +353,68 @@ end
...
+
+#     for (_, spinfo) in spinfos
+#         iszero(spinfo.ub) && continue
+#         for (sp_var_id, sp_var_val) in spinfo.bestsol
+#             push!(sol_ids, sp_var_id)
+#             push!(sol_vals, sp_var_val)
+#         end
+#     end
+#     return sparsevec(var_ids, var_vals) .* sparsevec(sol_ids, sol_vals)
+# end
```
guimarqu, line 416:
```suggestion
```

PR #722, opened 2022-09-20
mv cat with partial sol from Algorithms to MOI_interface

guimarqu:


src/MathProg/solutions.jl
```diff
@@ -184,7 +184,7 @@ end
 function Base.show(io::IO, solution::DualSolution{M}) where {M}
     println(io, "Dual solution")
     for (constrid, value) in solution
-        println(io, "| ", getname(getmodel(solution), constrid), " = ", value)
+        println(io, "| ", getname(getmodel(solution), constrid), " = ", value, "       (rhs = $(getcurrhs(getmodel(solution), constrid)))")
```
guimarqu, line 187:
```suggestion
        println(io, "| ", getname(getmodel(solution), constrid), " = ", value)
```

src/MathProg/solutions.jl
```diff
@@ -195,7 +195,7 @@ end
 function Base.show(io::IO, solution::PrimalSolution{M}) where {M}
     println(io, "Primal solution")
     for (varid, value) in solution
-        println(io, "| ", getname(getmodel(solution), varid), " = ", value)
+        println(io, "| ", getname(getmodel(solution), varid), " = ", value, "       (cost = $(getcurcost(getmodel(solution), varid)))")
```
guimarqu, line 198:
```suggestion
        println(io, "| ", getname(getmodel(solution), varid), " = ", value)
```

PR #736, opened 2022-09-30
Adding an integration test for pricing stages

artalvpes:
I started this to reproduce a bug but there was no bug. I was not using the api correctly. These tests also serve as examples of api use.


guimarqu:
Thanks Artur! I think it should be moved to documentation. I'll do it later.

PR #745, opened 2022-11-03
Add reformulation parser

### LINKED ISSUE #702: Model parser for tests
We need to build a simple formulation or a reformulation for several algorithms tests.
However, even if the formulation and the reformulation are simple, they are difficult to write and read. See for instance https://github.com/atoptima/Coluna.jl/blob/2e24949f459b00d47eddebb81ae3cdabbad0c1aa/test/unit/Algorithm/colgen.jl#L1-L40

In MathOptInterface.jl, they created a parser: https://github.com/jump-dev/MathOptInterface.jl/blob/b34036eabd91affcfd2b9eb26df115d59da800e2/src/Utilities/parser.jl

It would be nice to create something on top of their parser to create a reformulation.
For instance (syntax should be adapted to MOI parser if we decide to build on top of it):

```
"""
    master:
        min x + y + z1 + z2
        st x + y + z1 + 2*z2  >= 1
        y >= 2

    dw_sp:
        x + z1 <= 3

    dw_sp:
        x + 3z2 <= 2

    representatives:
        x
"""
```
where `y` is a master variable, `x` a dw sp representative, `z1` a variable of the first subproblem, and `z2` a variable of the second subproblem.

Advantage: test easier to understand
Disadvantage: it seems to be quite a lot of work & we have to make sure there is no bug in the parser.

eduardovegas:
Closes #702 .

### DIFF
--- a/test/integration/parser.jl
+++ b/test/integration/parser.jl
@@ -6,11 +6,11 @@ struct AuxiliaryConstrInfo
 end
 
 function get_vars_info(form::CL.Formulation)
-    names = []
-    kinds = []
-    duties = []
-    costs = []
-    bounds = []
+    names = String[]
+    kinds = ClMP.VarKind[]
+    duties = ClMP.Duty{ClMP.Variable}[]
+    costs = Float64[]
+    bounds = Tuple{Float64,Float64}[]
     for (varid, var) in CL.getvars(form)
         push!(names, CL.getname(form, var))
         push!(kinds, CL.getperenkind(form, var))
@@ -25,7 +25,7 @@ function get_constrs_info(form::CL.Formulation)
     infos = AuxiliaryConstrInfo[]
     coeff_matrix = CL.getcoefmatrix(form)
     for (constrid, constr) in CL.getconstrs(form)
-        coeffs = []
+        coeffs = Tuple{String,Float64}[]
         for (varid, coeff) in @view coeff_matrix[constrid, :]
             push!(coeffs, (CL.getname(form, varid), coeff))
         end
--- a/test/integration/parser.jl
+++ b/test/integration/parser.jl
@@ -51,7 +51,7 @@ end
                 y1 >= 1
                 1 <= y2
         """
-        @test_throws ErrorException CL.reformfromstring(s)
+        @test_throws ErrorException reformfromstring(s)
 
         # OF not defined
         s = """
@@ -70,7 +70,7 @@ end
                 y1 >= 1
                 1 <= y2
         """
-        @test_throws ErrorException CL.reformfromstring(s)
+        @test_throws ErrorException reformfromstring(s)
     end
 
     @testset "no subproblems" begin
@@ -90,7 +90,7 @@ end
             bounds
                 y <= 10
         """
-        env, master, subproblems, constraints = CL.reformfromstring(s)
+        env, master, subproblems, constraints = reformfromstring(s)
 
         @test CL.getobjsense(master) == CL.MaxSense
 
@@ -121,7 +121,7 @@ end
             bounds
                 y <= 10
         """
-        @test_throws ErrorException CL.reformfromstring(s)
+        @test_throws ErrorException reformfromstring(s)
     end
 
     @testset "variables not defined" begin
@@ -144,7 +144,7 @@ end
             bounds
                 x >= 10
         """
-        @test_throws ErrorException CL.reformfromstring(s)
+        @test_throws ErrorException reformfromstring(s)
 
         # master variable not present in OF
         s = """
@@ -159,7 +159,7 @@ end
             bounds
                 x, y >= 5
         """
-        @test_throws ErrorException CL.reformfromstring(s)
+        @test_throws ErrorException reformfromstring(s)
 
         # variable in OF with no duty and kind defined
         s = """
@@ -178,7 +178,7 @@ end
             bounds
                 2 <= x, w <= 10
         """
-        @test_throws ErrorException CL.reformfromstring(s)
+        @test_throws ErrorException reformfromstring(s)
 
         # variable in constraint with no duty and kind defined
         s = """
@@ -197,7 +197,7 @@ end
             bounds
                 2 <= x, w <= 10
         """
-        @test_throws ErrorException CL.reformfromstring(s)
+        @test_throws ErrorException reformfromstring(s)
 
         # subproblem variable with no duty and kind defined
         s = """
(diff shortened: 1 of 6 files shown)


guimarqu:
Do you think you can move this outside Coluna? I think it would be nice to keep it in the tests. I don't know if we can have a kind of `TestUtils` module in the test folder or something else similar.

eduardovegas:
Would it be better if the parser file in src/ was moved to test/, and add include("parser.jl") in ColunaTests module? Because creating a submodule I'd have to define some constants that are already in ColunaTests and export the functions from the new module.

guimarqu:
> Would it be better if the parser file in src/ was moved to test/, and add include("parser.jl") in ColunaTests module? Because creating a submodule I'd have to define some constants that are already in ColunaTests and export the functions from the new module.

Sounds good :)

test/integration/parser.jl
```diff
@@ -0,0 +1,380 @@
...
+    sense::CL.ConstrSense
+    rhs::Float64
+end
+
+function get_vars_info(form::CL.Formulation)
+    names = []
+    kinds = []
+    duties = []
+    costs = []
+    bounds = []
```
guimarqu, line 13:
```suggestion
    names = String[]
    kinds = ClMP.VarKind[]
    duties = ClMP.Duty{ClMP.Variable}[]
    costs = Float64[]
    bounds = Tuple{Float64,Float64}[]
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+    end
+    return vars
+end
+
+function _read_expression(l::AbstractString)
+    line = _strip_line(l)
+    vars = Dict{String, Float64}()
+    first_m = match(Regex("^([+-])?($coeff_re)?\\*?(\\w+)(.*)"), line) # first element of expr
+    if first_m !== nothing
+        sign = first_m[1] === nothing ? "+" : first_m[1] # has a sign
```
guimarqu, line 134:
```suggestion
        sign = isnothing(first_m[1]) ? "+" : first_m[1] # has a sign
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+        coeff = first_m[2] === nothing ? "1" : first_m[2] # has a coefficient
```
guimarqu, line 135:
```suggestion
        coeff = isnothing(first_m[2]) ? "1" : first_m[2] # has a coefficient
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+        cost = parse(Float64, string(sign, coeff))
+        vars[String(first_m[4])] = cost
+        for m in eachmatch(Regex("([+-])($coeff_re)?\\*?(\\w+)"), first_m[5]) # rest of the elements
+            coeff = m[2] === nothing ? "1" : m[2]
```
guimarqu, line 139:
```suggestion
            coeff = isnothing(m[2]) ? "1" : m[2]
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+            vars[String(m[4])] = cost
+        end
+    end
+    return ExprCache(vars)
+end
+
+function _read_constraint(l::AbstractString)
+    line = _strip_line(l)
+    m = match(Regex("(.+)(>=|<=|==)($coeff_re)"), line)
+    if m !== nothing
```
guimarqu, line 150:
```suggestion
    if !isnothing(m)
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+    bound1, bound2 = ("","")
+    m = match(r, line)
+    if m !== nothing
+        vars = _get_vars_list(m[4]) # separate variables as "x_1, x_2" into a list [x_1, x_2]
+        if m[1] !== nothing # has lower bound (nb <= var) or upper bound (nb >= var)
+            bound1 = String(m[2])
+        end
+        if m[5] !== nothing # has upper bound (var <= nb) or lower bound (var >= nb)
+            bound2 = String(m[6])
+        end
```
guimarqu, line 179:
```suggestion
    if !isnothing(m)
        vars = _get_vars_list(m[4]) # separate variables as "x_1, x_2" into a list [x_1, x_2]
        if !isnothing(m[1]) # has lower bound (nb <= var) or upper bound (nb >= var)
            bound1 = String(m[2])
        end
        if !isnothing(m[5]) # has upper bound (var <= nb) or lower bound (var >= nb)
            bound2 = String(m[6])
        end
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+
+function read_master!(sense::Type{<:ClB.AbstractSense}, cache::ReadCache, line::AbstractString)
+    obj = _read_expression(line)
+    cache.master.sense = sense
+    cache.master.objective = obj
+end
+
+function read_master!(::Val{:constraints}, cache::ReadCache, line::AbstractString)
+    constr = _read_constraint(line)
+    if constr !== nothing
```
guimarqu, line 192:
```suggestion
    if !isnothing(constr)
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+        push!(cache.master.constraints, constr)
+    end
+end
+
+read_master!(::Any, cache::ReadCache, line::AbstractString) = nothing
+
+function read_subproblem!(cache::ReadCache, line::AbstractString, nb_sp::Int64)
+    constr = _read_constraint(line)
+    if constr !== nothing
```
guimarqu, line 201:
```suggestion
    if !isnothing(constr)
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+    #create subproblems
+    subproblems = []
+    all_spvars = Dict{String, ClMP.Variable}()
+    for (_, sp) in cache.subproblems
+        spform = nothing
+        for varid in sp.varids
+            if haskey(cache.variables, varid)
+                var = cache.variables[varid]
+                if var.duty <= ClMP.DwSpPricingVar || var.duty <= ClMP.MasterRepPricingVar
+                    if spform === nothing
```
guimarqu, line 266:
```suggestion
                    if isnothing(spform)
```

test/parser.jl
```diff
@@ -0,0 +1,401 @@
...
+                end
+            else
+                error("Variable $varid duty and/or kind not defined")
+            end
+        end
+        c = ClMP.setconstr!(master, "c$i", constr_duty; rhs = constr.rhs, sense = constr.sense, members = members)
+        push!(constraints, c)
+        i += 1
+    end
+    #create subproblems constraints in master
```
guimarqu, line 343: This is a very long method. Is it possible to isolate each step in a specific method?

guimarqu (changes requested):
This is a nice PR. Some minor style issues to address.

test/parser.jl
eduardovegas: Yes!

guimarqu (approved):
Thanks @eduardovegas!

PR #746, opened 2022-11-16
Improve relax callback

artalvpes:
This branch introduces two independent features that are related to the same application of improving relaxations: a callback function to be executed before every cut generation and a mechanism to pass information from nodes to their descendants.


guimarqu (approved):
Merging into a "dev" branch because we could use regular storage but the code needs improvement.

PR #824, opened 2023-04-20
Test of `ReducedCostCalculationHelper` with non-robust cuts

enjoyUrFruits:
Expected redcosts differ from redcosts computed with ReducedCostsCalculationHelper, even when removing the r1c constraint. There may be a problem in the way expected redcosts are computed.


test/unit/ColGen/colgen_default.jl
```diff
@@ -1379,3 +1321,148 @@ function test_colgen()
...
+        1 0 0 0 0;
+        0 1 0 0 0;
+        0 0 1 0 0;
+        0 0 0 1 0;
+        0 0 0 0 1;
+        0 0 0 0 0 
+    ]
+
+
+    expected_redcosts = costs - transpose(A)*[2.0, 8.0, 1.0, 3.0, 9.0, 4.0]
```
guimarqu, line 1461: `expected_redcosts` calculation is fine.

test/unit/ColGen/colgen_default.jl
```diff
@@ -1379,3 +1321,148 @@ function test_colgen()
...
+        3.0,
+        11.0,
+        7.0
+    ]
+
+    constrids = [id for (id,_) in ClA.getconstrs(master)]
+    @show typeof(constrids)
+
+    dual_sol = DynamicSparseArrays.sparsevec(
+        constrids,
```
guimarqu, line 1447: Are you sure constraint ids are in the right order?

guimarqu (approved):

PR #897, opened 2023-05-12
[adv tuto] add comparison of different cuts on a larger instance

enjoyUrFruits:


docs/src/start/advanced-demo.jl
```diff
@@ -815,4 +804,104 @@ routes_costs = Dict(
...
+locations  = vcat(facilities, customers)
+nb_customers = length(customers)
+nb_facilities = length(facilities)
+positions = 1:nb_positions;
+
+routes_per_facility = Dict(
+    j => best_route_forall_cust_subsets(arc_costs, customers, j, nb_positions) for j in facilities
+);
+
+# We set maxnumnodes to zero to optimize only the root node:
```
guimarqu, line 854:
```suggestion
# We set `maxnumnodes` to zero to optimize only the root node:
```

docs/src/start/advanced-demo.jl
```diff
@@ -815,4 +804,104 @@ routes_costs = Dict(
...
 @benders_decomposition(model, dec, axis)
 JuMP.optimize!(model)
-## @show objective_value(model)
+@show objective_value(model)
+
+
+
+
+
+# ## Comparison of the results obtained on a larger instance
```
guimarqu, line 813:
```suggestion
# ## Example of comparison of the dual bounds obtained on a larger instance.
```

guimarqu (approved):

PR #933, opened 2023-06-14
Test primal solution used to compute subgradient in alpha self-adjusting procedure

guimarqu:


test/unit/ColGen/colgen_stabilization.jl
```diff
@@ -15,8 +15,101 @@ function test_misprice_schedule()
...
+    @test primal_sol[vids["x_12"]] == 2.0 * sp1_lb
+    @test primal_sol[vids["x_13"]] == 3.0 * sp1_lb
+    @test primal_sol[vids["x_21"]] == 4.0 * sp2_ub
+    @test primal_sol[vids["x_22"]] == 4.0 * sp2_ub
+    @test primal_sol[vids["x_23"]] == 5.0 * sp2_ub
+    @test primal_sol[vids["y1"]] == 7.0 * sp1_lb
+    @test primal_sol[vids["y2"]] == 10.0 * sp2_ub
+    @test primal_sol[vids["s1"]] == 1.0 * sp1_lb
+    @test primal_sol[vids["s2"]] == 1.0 * sp2_ub
+    @test primal_sol[vids["z1"]] == 0.0 # TODO: not sure about this test
```
guimarqu, line 110: @rrsadykov  I'm don't know what value should take pure master variables in the "fake primal solution" that we use to compute the subgradient in the dynamic alpha procedure. Shoud the value be equal to the variable lower bound or upper bound depending on the sign of the variable in the objective?

guimarqu: Note that in 0.5.3 the value is always 0.

rrsadykov: Subgradient is equal to the vector of RHS minus the coefficient matrix times the total solution of pricing suproblems under the bounding constraints. There is one (trivial) pricing subproblem for each pure master variable. Solution to this subproblem is equal to the master variable upper bound if its reduced cost is negative (and not the cost in the objective), and equal to the master variable lower bound if its reduced cost is positive. Reduced cost of a pure master variable is calculated based on the smoothed dual solution.

PR #939, opened 2023-06-15
Test Benders flow for infeasible/unbounded

enjoyUrFruits:
Two tests in ```benders_flow.jl```

- benders_flow_unbounded_master ; I check that we enter ```treat_unbounded_master_problem_case!``` method, the test passed
- benders_flow_infeasible_sp ; I create a formulation with some 1st level variables fixed so that the sp is infeasible. I was expected to enter ```treat_infeasible_separation_problem_case!``` but this is not the case. But maybe it is not the expected behaviour given my formulation ?

### DIFF
--- a/test/unit/Benders/benders_flow_extra.jl
+++ b/test/unit/Benders/benders_flow_extra.jl
@@ -78,7 +78,7 @@ function benders_flow_form_unbounded_master()
     result = Coluna.Benders.run_benders_iteration!(ctx, nothing, env, nothing)
     @test ctx.flag_unbounded_master == true
 end
-register!(unit_tests, "benders_default", benders_flow_form_unbounded_master, f=true)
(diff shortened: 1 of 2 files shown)


test/unit/Benders/benders_flow_extra.jl
```diff
@@ -152,11 +147,12 @@ function benders_infeasible_sp()
     for (_, sp) in Coluna.MathProg.get_benders_sep_sps(reform)
         sp.optimizers = Coluna.MathProg.AbstractOptimizer[] # dirty
         ClMP.push_optimizer!(sp, () -> ClA.MoiOptimizer(GLPK.Optimizer()))
+        Coluna.Algorithm._deactivate_art_vars(sp)
```
guimarqu, line 150: ping @enjoyUrFruits

PR #947, opened 2023-06-22
Improve column generation dev doc (1)

guimarqu:


docs/src/api/colgen.md
```diff
@@ -35,15 +103,23 @@ Coluna.ColGen.setup_context!
...
 
-## Stages
+### Stages
 
-A stage is a set of consecutive iterations in which we use a given pricing solver. 
-The goal is to solve the pricing subproblem with very fast heuristic solvers first and then
-switch to a "more exact" solver when a given condition is met. The last stage generally uses
-an exact solver.
+A stage is a set of consecutive iterations in which we use a given pricing solver.
+The goal is to solve the pricing subproblem with very fast heuristic solvers at the first stage
```
enjoyUrFruits, line 109: This is a bit ambiguous ; it makes me believe that multi-stages is only composed of 2 stages:
- a first stage with several heuristic solvers
- a second (and last) stage with an exact solver
It is not the case, right ?

```suggestion
The aim is to speed up the resolution of the pricing problem by first using an approximate but fast pricing algorithm and then switching to increasingly less heuristic algorithms until the last stage where an exact solver is used.
```

docs/src/api/colgen.md
```diff
@@ -55,13 +131,47 @@ Coluna.ColGen.stage_id
 Coluna.ColGen.is_exact_stage
 ```
 
-The default implementation of the stages is as follows.
-
-```@docs
-Coluna.ColGen.ColGenStageIterator
+### Column generation iteration
+
+These are the main steps of a column generation iteration without stabilization.
```
enjoyUrFruits, line 136: Add link to stabilization doc (when it exists ofc :) )

docs/src/api/colgen.md
```diff
@@ -146,26 +285,54 @@ Coluna.ColGen.get_dual_bound
...
+
+Go back to the [column generation iteration overview](#Column-generation-iteration).
+
+#### Iteration output
+
+Lorem ipsum.
+
+Go back to the [column generation iteration overview](#Column-generation-iteration).
+
+### Stabilization
```
enjoyUrFruits, line 335: TODO: add an introduction to what stabilization does

enjoyUrFruits (approved):

enjoyUrFruits:
I think it could be relevant to add a dedicated paragraph (or even a section, it should be visible) about how to read the output of column generation. I think the only example of output interpretation is in Manual->Algorithms->Conquer Algorithm. It is quite hard to find while it may be one of the first information a user wants to get.

PR #950, opened 2023-06-22
Fix lagrangian bound calculation in column generation

guimarqu:
- Maintain the incumbent dual bound during the whole column generation phase and check the convergence with this incumbent dual bound (so we are sure the lagrangian bound is well calculated at each iteration)
- Fix computation of the valid dual bound


src/Algorithm/colgen/default.jl
```diff
@@ -682,20 +687,59 @@ end
...
-function ColGen.compute_dual_bound(ctx::ColGenContext, phase, sp_dbs, master_dual_sol)
-    master_lp_obj_val = getvalue(master_dual_sol)
-    sp_contrib = mapreduce(((id, val),) -> val, +, sp_dbs)
-    convexity_contrib = _convexity_contrib(ctx, master_dual_sol)
+function ColGen.compute_dual_bound(ctx::ColGenContext, phase, sp_dbs, generated_columns, master_dual_sol)
+    sc = ColGen.is_minimization(ctx) ? 1 : -1
+    master_lp_obj_val = if ctx.stabilization
+        sc * (transpose(master_dual_sol) * ctx.subgradient_helper.a)
+    else
+        getvalue(master_dual_sol) #- _convexity_contrib(ctx, master_dual_sol) # TODO: understand what's going on here.
```
guimarqu, line 735: @rrsadykov @artalvpes if you find where I remove the convexity cnostraints contribution, that would be very nice. I'm blind.

- `getvalue(master_dual_sol)` should be equal to the master lp obj
- dual bound of each subproblem contains the cost of subproblem variables only.

In theory, we should remove the contribution of convexity constraints... but if I do so, it doesn't work.

guimarqu: I'll compute the dual bound from scratch using the text-book formula tomorrow to understand what's happening.

guimarqu: I found bug 😅

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.6.2"
 
 [deps]
 BlockDecomposition = "6cde8614-403a-11e9-12f1-c10d0f0caca0"
+ColunaDemos = "a54e61d4-7723-11e9-2469-af255fcaa246"
```
guimarqu, line 8:
```suggestion
```

revise.sh
```diff
@@ -1,6 +1,6 @@
 #/bin/bash
 
-alias julia='/Applications/Julia-1.8.app/Contents/Resources/julia/bin/julia'
+alias julia='/Applications/Julia-1.9.app/Contents/Resources/julia/bin/julia'
```
guimarqu, line 3:
```suggestion
alias julia='/Applications/Julia-1.8.app/Contents/Resources/julia/bin/julia'
```

PR #953, opened 2023-06-23
Attempt to update man/algorithms

enjoyUrFruits:
This is a suggestion on how the manual section "Algorithms" could be transformed. I change the name into "Built-in algorithms for Branch-and-Bound", write an intro on Branch-and-Bound and  ```Algorithm.TreeSearchAlgorithm``` and create several subsections where I organize the different pieces of doc provided by Coluna. This is may be too exhaustive ? 

PS: I did not find why, but it looks that the references to the different algorithms are broken in my new files, I did not suceed to fix it.


docs/src/man/builtin-algo.md
```diff
@@ -0,0 +1,82 @@
...
+
+In this 2-dimensional example, a feasible solution of the MIP has been found in subspace/node B (circled in blue). The dual bounds of all subspaces have been computed (indicated by blue stars on the polyhedron). We notice that dual bounds of subspaces A and D are worse than this current feasible solution (w.r.t. the objective function whose direction is indicated by the gray arrow). Thus, we can avoir exploring subspaces A and D, and nodes A and D can be pruned. 
+
+In order to speed up the algorithm, it could be helpful to apply some decompositions s.t. Dantzig-Wolfe or Benders ahead of the Branch-and-Bound tree. However, these decompositions lead to the creation of an exponential number of variables in the formulation. To avoid spending too much time optimizing each node, the alternative is to solve the linear relaxation at each node using Cut, and/or Column Generation (for Dantzig-Wolfe decomposed problems only). This approach leads to algorithms known as Branch-and-Cut (Branch-and-Bound + Cut Generation), Branch-and-Price (Branch-and-Bound + Column Generation) and Branch-and-Cut-and-Price (Branch-and-Bound + Cut and Column Generation).
```
guimarqu, line 15: Il would start like this:

```suggestion
Coluna provides a generic Branch-and-Bound algorithm whose three main elements can be easily modified::
- the conquer strategy: evaluation of the problem at a node of the branch-and-bound tree
- the branching strategy: ...
- the explore strategy: the evaluation order of your nodes 
```

Then you can talk about the conquer strategy that can use column generation if you applied DW decomposition, cut generation if Benders

docs/src/man/builtin-algo.md
```diff
@@ -0,0 +1,82 @@
...
+The main loop used for the implementation of such algorithms is implemented by ```Algorithm.TreeSearchAlgorithm```. It is highly customizable so that the user can specify its own strategy to optimize each node (conquer) and to explore the tree:
+
+```@docs
+Algorithm.TreeSearchAlgorithm
+```
+
+In the following sections, we present the global setup of ```Algorithm.TreeSearchAlgorithm```. First, we describe the construction and the exploration of the search tree. Then, we "zoom" on the tree to describe the setup of Cut and Column Generation at each node.
+
+
+## Manage the branching phase
```
guimarqu, line 26: Then, I think we should just give a link to the algorithms available for the conquer and the branching strategies.

enjoyUrFruits: I think it is the case (modulo the broken links)

PR #956, opened 2023-06-26
Update man/callbacks

enjoyUrFruits:
I just rename the section to "User-defined callbacks" to underline the difference with the other algorithms (i.e. the built-in algorithms) and add some TODOs.

### DIFF
--- a/docs/src/man/callbacks.md
+++ b/docs/src/man/callbacks.md
@@ -48,7 +48,3 @@ and [facultative cut separation](https://jump.dev/JuMP.jl/stable/tutorials/Mixed
 in the JuMP documentation.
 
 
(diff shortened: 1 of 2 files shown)


docs/src/man/callbacks.md
```diff
@@ -48,3 +48,7 @@ and [facultative cut separation](https://jump.dev/JuMP.jl/stable/tutorials/Mixed
 in the JuMP documentation.
 
 
+## Branching callbacks
+
+TODO
+
```
guimarqu, line 54: There is no branching callback at the moment

guimarqu, line 54:
```suggestion
```

PR #962, opened 2023-06-27
Update tuto custom data

enjoyUrFruits:
I removed the identical subproblems from the tutorial on custom data -- but now the tuto is broken. I don't get the good solution at the end of the optimization anymore. Maybe I forgot to update something in the pricing callback ?


guimarqu:
Maybe we should use another example for the custom data tutorial. I'm afraid the current example needs too much code.

guimarqu (approved):

PR #966, opened 2023-06-28
add formulas for man -> benders

enjoyUrFruits:


guimarqu:
I already put the reformulation here: https://atoptima.github.io/Coluna.jl/dev/api/benders/#Reformulation

Can you make your own formulas consistent with the Benders API doc pls?

PR #972, opened 2023-07-03
init documentation for user-supplied parameters checkings

enjoyUrFruits:
I start to write a description of what could be put in place to perform checks of user-supplied algorithms and parameters in ```src/api/checks_algos.md```


guimarqu (commented):
Ok but I think we should name the page "Algorithms and initialization"

guimarqu (approved):

PR #988, opened 2023-07-15
Fix the definition of the NodeFinalizer algorithm

artalvpes:
It was not defined as an AbstractAlgorithm. Why didn't the tests get this bug?


guimarqu:
This test in is in `extra_e2e`, a test set I run before a new release.

PR #989, opened 2023-07-15
Stop restoring the subproblem solver parameters to avoid spurious prints

artalvpes:
Restoring the solver parameter value was setting the output flag to 1, causing the solver to repeatedly print messages about parameter setting.


guimarqu:
Close in favor of #1006.

PR #998, opened 2023-07-20
Correction of run_divide routine and update of treesearch tests

najaverzat:
I removed the condition ```nodestatus == OPTIMAL``` from routine run_divide. I updated the tests of treesearch. The second test should check the behaviour of the algorithm when the gap is closed. It should work after the refactoring of the checks before conquer run.


src/Algorithm/treesearch.jl
```diff
@@ -131,7 +131,8 @@ Performs operations after the divide algorithm when the current node is finally
 function run_divide(divide_input)
     conquer_opt_state = Branching.get_conquer_opt_state(divide_input)
     nodestatus = getterminationstatus(conquer_opt_state)
-    return !(nodestatus == OPTIMAL || nodestatus == INFEASIBLE || ip_gap_closed(conquer_opt_state))             
+    @show nodestatus
```
guimarqu, line 134:
```suggestion
```

src/Algorithm/treesearch.jl
```diff
@@ -156,6 +157,8 @@ function TreeSearch.children(space::AbstractColunaSearchSpace, current::TreeSear
     # if run_divide returns false, the divide is not run and the node is pruned.
     if run_divide(divide_input)
         branches = run!(divide_alg, env, reform, divide_input)   
+    else
+        println("\e[31m no children generated \e[00m")
```
guimarqu, line 161:
```suggestion
```

PR #1002, opened 2023-07-20
Update tests treesearch

najaverzat:
We modified run_divide and run_conquer. Thus the tests have been updated and also corrected. They are still not integrated as "real tests" but they work.

However some tests are now broken. I suspect my update of children in ```treesearch.jl``` to be the cause ; especially the way I exit the method if run_conquer is false.


guimarqu:
I removed the `run_conquer` condition in the implementation of `TreeSearch.children` for the `AbstractColunaSearchSpace`. Indeed, we need to run the `node_is_leaf` method when the node is pruned, and the `conquer_output` is required. We expect the conquer algorithm to stop as soon as it finds the node pruned.

Other tests didn't break. Are your tests running here?

najaverzat:
My tests are not yet integrated as real tests so they don't run. I have run them manually and yes, they are broken now because we enter the run! of conquer at unexpected moments. Plus, as we removed the exit condition that was previously in run_colcutgen_conquer, I guess we are making too much computations in other tests -- they don't fail because it is not false but it is suboptimal.

src/Algorithm/treesearch.jl
```diff
@@ -166,30 +152,25 @@ function TreeSearch.children(space::AbstractColunaSearchSpace, current::TreeSear
...
+    conquer_output = run!(conquer_alg, env, reform, conquer_input)
     after_conquer!(space, current, conquer_output) # callback to do some operations after the conquer.
     # Build the divide input from the conquer output
     divide_alg = get_divide(space)
     divide_input = get_input(divide_alg, space, current, conquer_output)
     branches = nothing
-    # if `run_divide` returns false, the divide is not run and the node is pruned.
-    if run_divide(divide_input)
+    # if `run_divide` returns false, the divide is not run and the node is pruned. 
+    # TODO?: node_is_pruned and node_is_leaf seem redondant. We could maybe avoid to have two different methods by overloading one (with conquer_output as extra parameter)
```
guimarqu, line 172:
```suggestion
```

src/Algorithm/treesearch/branch_and_bound.jl
```diff
@@ -208,6 +207,36 @@ function after_conquer!(space::BaBSearchSpace, current, conquer_output)
...
+    # TODO: improve ?
+    # Condition 1: IP Gap is closed. Abort treatment.
+    # Condition 2: in the case the conquer was already run (in strong branching),
+    # Condition 3: make sure the node has not been proven infeasible.
+    # we still need to update the node IP primal bound before exiting 
+    # (to possibly avoid branching)
+    run_conquer = !ip_gap_closed(conquer_input.node_state, rtol = space.opt_rtol, atol = space.opt_atol)
+    run_conquer = run_conquer && getterminationstatus(conquer_input.node_state) != INFEASIBLE
+    return run_conquer
+end
```
guimarqu, line 220:
```suggestion
```

src/Algorithm/treesearch.jl
```diff
@@ -134,29 +134,15 @@ Performs operations after the divide algorithm when the current node is finally
...
-    node_state = OptimizationState(
-        getmaster(space.reformulation);
-        ip_dual_bound = current.ip_dual_bound
-    )
-    run_conquer = !ip_gap_closed(node_state, rtol = space.opt_rtol, atol = space.opt_atol)
-    run_conquer = run_conquer || !current.conquerwasrun
-    run_conquer = run_conquer && getterminationstatus(node_state) != INFEASIBLE
-    return run_conquer
-end
+@mustimplement "ColunaSearchSpace" run_conquer(sp::AbstractColunaSearchSpace, conquer_input, node) = nothing
```
guimarqu, line 139:
```suggestion
```

src/Algorithm/treesearch/branch_and_bound.jl
```diff
@@ -208,6 +207,36 @@ function after_conquer!(space::BaBSearchSpace, current, conquer_output)
...
+function is_pruned(space::BaBSearchSpace, current::Node)
+    return MathProg.gap_closed(get_ip_primal_bound(space.optstate),
+                               current.ip_dual_bound,
+                               atol = space.opt_atol,
+                               rtol = space.opt_rtol)
+end
+
+function node_is_pruned(space::BaBSearchSpace, current::Node)
+    leaves_status = space.leaves_status
+    leaves_status.infeasible = false # we closed the gap so the problem is feasible. Natacha: I would have say: we have a primal bound, so a primal solution, so the problem is feasible (even if the current leaf is infeasible).
```
guimarqu, line 231:
```suggestion
    leaves_status.infeasible = false # We have a primal bound, so a primal solution, and we closed the gap, so the original problem is feasible. 
```

src/Algorithm/treesearch/branch_and_bound.jl
```diff
@@ -232,13 +261,23 @@ function get_input(::AbstractConquerAlgorithm, space::BaBSearchSpace, current::N
...
 function get_input(::AlgoAPI.AbstractDivideAlgorithm, space::BaBSearchSpace, node::Node, conquer_output)
     return DivideInputFromBaB(node.depth, conquer_output, node.records)
 end
 
 number_of_children(divide_output::DivideOutput) = length(divide_output.children)
 
 function node_is_leaf(space::AbstractColunaSearchSpace, current::Node, conquer_output::OptimizationState)
+    println("\e[41m node is leaf \e[00m")
+    @show getterminationstatus(conquer_output)
+
```
guimarqu, line 280:
```suggestion
```

PR #1003, opened 2023-07-20
Fix the non robust cuts contribution when optimizing the subproblems

artalvpes:
This contribution is essential to correctly identify the columns that should be added to the restricted master.


guimarqu:
Hey Artur,

Thanks for the report.

We should discuss this because we expect the user to compute the contribution of non-robust-cuts in the pricing callback.

There is an example in the documentation: https://atoptima.github.io/Coluna.jl/stable/start/advanced_demo/#Strengthening-the-master-with-valid-inequalities-on-the-column-generation-variables-(so-called-%22non-robust%22-cuts)

guimarqu:
UPDATE: we don't use the same dual solution in the pricing subproblem and when computing the real reduced cost when using stabilization so my previous comment is wrong for the case where stabilization is active and this fix is necessary.

PR #1005, opened 2023-07-21
create subroutine gap_closed in bounds.jl

najaverzat:
We create a subroutine `gap_closed` that takes as two bounds as arguments  instead of an optimization state (as in `_ip_gap_closed`). ```_ip_gap_closed``` calls this subroutine now.

### DIFF
--- a/src/MathProg/bounds.jl
+++ b/src/MathProg/bounds.jl
@@ -86,6 +86,7 @@ function ObjValues(
     return ov
 end
 
+
 ## Gaps
 _ip_gap(ov::ObjValues) = gap(ov.ip_primal_bound, ov.ip_dual_bound)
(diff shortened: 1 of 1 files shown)


src/MathProg/bounds.jl
```diff
@@ -86,18 +86,26 @@ function ObjValues(
...
+function gap_closed(
+    pb::Bound, db::Bound; atol = Coluna.DEF_OPTIMALITY_ATOL, rtol = Coluna.DEF_OPTIMALITY_RTOL
+)
+    return gap(pb, db) <= 0 || _gap_closed(
+        pb.value, db.value, atol = atol, rtol = rtol
     )
 end
 
+
 function _lp_gap_closed(
```
guimarqu, line 109: You can use  gap_closed in _lp_gap_closed also

PR #1038, opened 2023-08-25
Correction for the bug in testing the integrality of projected solution

rrsadykov:


guimarqu (approved):
We'll have the same bug when using multiple subproblems with a high upper multiplicity limit (especially the case when we use the same column `n` times). The numerical error from the subsolver will be multiplied by the upper multiplicity (worst case) but tolerance will stay the same.

PR #1052, opened 2023-09-07
limited discrepancy search explore strategy

guimarqu:


rrsadykov (commented):
I am not sure that LDS as it is implemented is a proper explore strategy. For me, explore strategy is for more about in which order we explore generated children, and not about filtering the generated children.  The original LDS was proposed in http://cse.unl.edu/~choueiry/Documents/LDS.pdf as a complete search, but with different explore order of nodes. 

Also, the LDS explore strategy as it is implemented now is not that much useful, as it just truncates the number of children generated, if we generate too much. I think we should already generate exactly the number of children we need  in diving, as every  generated child is not cheap (we do preprocessing for each generated child).

guimarqu:
I agree, I'll move the code so we can qualify it as a tree search algorithm. It won't the implemantation of the lds though. We'll have to figure out how we can transmit the number of exepected children to the divide algorithm.

Issue #1061, opened 2023-09-13
Issues with variable fixing


rrsadykov:
There are some issus with variable fixing
- First, columns should never be fixed, as we cannot impose an upper bound on a column (it is too expensive to take into account the dual value associated with the upper bound in the pricing). So the `MasterColumnsRecord` and `unit_master_columns_record()` test should be modified. For me, we should only keep the set of active columns in `MasterColumnsRecord` without anything else.
- Second, there is now ambiguity between fixing a variable and deactivating it. For the moment, fixing deactivates the variable, but just deactivation does not fixes the variable. When variable is deactivated but not fixed, what does this mean? For me, this means that lower and upper bound are fixed to zero. Thus may be we should set bounds to zero when we just deactivate the variable? Because of this ambiguity I think that recording and restoring `StaticVarConstrUnit` is not done completely correctly, we should review it.

guimarqu:
About the second point, 

For all variables, fixing => deactivation.

Deactivation means the variable is not part of the formulation anymore.
So when you deactivate, we ask the subsolvers to not consider the variable anymore.
It can mean you fix it to 0 but this is perhaps not the case everywhere. This is something to investigate.

Deactivation is mainly used to restore the state of a formulation when starting to evaluate a tree search node.
Also when cleaning up the master in column generation (currently unavailable). In the latter case, activating the master column variable when finding the same subproblem solution can be interpreted as an unfix.

guimarqu:
So for the partial solution, I suggest to keep the fixed variables field :

https://github.com/atoptima/Coluna.jl/blob/b788c69212fc56442eb6e650e98ef7454b9aab97/src/MathProg/manager.jl#L50

but we indeed also need a new field to store the fixed lower bound on the variable.

We can have it in the FormulationManager, together with the set of fixed variables.
I'm open to a name better than "fixed lower bounds", or a name from the literature.
At the moment, we'll fix lower bounds only in the preprocessing.

rrsadykov:
Ok for me. May call this field `fixed_part` or `partial_value`

guimarqu:
go for partial_value

rrsadykov:
Or we can have one field `partial_solution::Dict{VarId,Float64}` in `FormulationManager` instead of two fields `partial_value` and `fixed_vars`

guimarqu:
I disagree because when the variable is fixed, we can deactivate it. This is not the case when it has a partial value. So I think we need to make a distinction.

PR #1064, opened 2023-09-14
Redefine partial solution

guimarqu:
ping @rrsadykov 

related to #1061

### DIFF
--- a/src/MathProg/MOIinterface.jl
+++ b/src/MathProg/MOIinterface.jl
@@ -246,13 +246,10 @@ function get_primal_solutions(form::F, optimizer::MoiOptimizer) where {F <: Form
             end
         end
         fixed_obj = 0.0
-        for var_id in getfixedvars(form)
-            fixed_val = getcurlb(form, var_id)
-            if abs(fixed_val) > Coluna.TOL
-                push!(solvars, var_id)
-                push!(solvals, fixed_val)
-                fixed_obj += getcurcost(form, var_id) * fixed_val
-            end
+        for (var_id, fixed_val) in getpartialsol(form)
+            push!(solvars, var_id)
+            push!(solvals, fixed_val)
+            fixed_obj += getcurcost(form, var_id) * fixed_val
         end
         solcost += fixed_obj
         push!(solutions, PrimalSolution(form, solvars, solvals, solcost, FEASIBLE_SOL))
@@ -322,10 +319,10 @@ function get_dual_solutions(form::F, optimizer::MoiOptimizer) where {F <: Formul
             end
         end
         fixed_obj = 0.0
-        for var_id in getfixedvars(form)
+        for (var_id, fixed_val) in getpartialsol(form)
             cost = getcurcost(form, var_id)
             if abs(cost) > Coluna.TOL
-                fixed_obj += cost * getcurlb(form, var_id)
+                fixed_obj += cost * fixed_val
             end
         end
         solcost += fixed_obj
--- a/src/MathProg/manager.jl
+++ b/src/MathProg/manager.jl
@@ -92,7 +92,7 @@ end
 function _add_partial_value!(m::FormulationManager, var::Variable, value)
     partial_value = get(m.partial_solution, var.id, 0.0)
     new_value = partial_value + value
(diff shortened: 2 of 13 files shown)


src/MathProg/manager.jl
```diff
@@ -45,9 +45,19 @@ Base.transpose(m::CoefficientMatrix) = transpose(m.matrix)
...
+
+    # When a variable is fixed, the variable is deactivated (not in the formulation anymore).
+    # Cost of the fixed variable is put in the objective constant.
     fixed_vars::Set{VarId}
+
+    # The partial value is a lower bound on the value of the variables in the solution.
+    # We remove this fixed part from the formulation and treat the variable like a classic
+    # non-negative one (>= 0).
+    # Cost of the partial value of the variable is put in the objective constant.
+    partial_values::Dict{VarId, Float64}
```
rrsadykov, line 59: For me, this is exactly partial solution. Why not call it `partial_solution`?

src/MathProg/manager.jl
```diff
@@ -45,9 +45,19 @@ Base.transpose(m::CoefficientMatrix) = transpose(m.matrix)
 mutable struct FormulationManager
     vars::Dict{VarId, Variable}
     constrs::Dict{ConstrId, Constraint}
-    objective_constant::Float64
     coefficients::ConstrVarMatrix # rows = constraints, cols = variables
+    objective_constant::Float64
+
+    # When a variable is fixed, the variable is deactivated (not in the formulation anymore).
+    # Cost of the fixed variable is put in the objective constant.
     fixed_vars::Set{VarId}
```
rrsadykov, line 53: If we have already `partial_values` (or `partial_solution`), I do not see any need in `fixed_vars`.  If variable's bounds are $[0,0]$, it can be just deactivated. If variable's bounds are equal and non-zero, for example $x\in[3,3]$, instead of fixing it, we can add $x=3$ to the partial solution, this decreases its bounds to  $[0,0]$, and we are in the first case (it can be just deactivated).

src/MathProg/manager.jl
guimarqu: We can call it a partial solution. I'll what you wrote in the other comment. I agree to remove fixed_vars (took one day to convince me :P)

src/MathProg/MOIinterface.jl
```diff
@@ -322,10 +319,10 @@ function get_dual_solutions(form::F, optimizer::MoiOptimizer) where {F <: Formul
             end
         end
         fixed_obj = 0.0
-        for var_id in getfixedvars(form)
+        for (var_id, fixed_val) in getpartialsol(form)
```
rrsadykov, line 322: Why we cannot just store the value of the partial solution so that it is not recomputed each time?

src/MathProg/manager.jl
```diff
@@ -79,17 +85,18 @@ function _addvar!(m::FormulationManager, var::Variable)
...
-    return
-end
-
-function _unfixvar!(m::FormulationManager, var::Variable)
-    delete!(m.fixed_vars, getid(var))
-    return
+function _add_partial_value!(m::FormulationManager, var::Variable, value)
+    partial_value = get(m.partial_solution, var.id, 0.0)
+    new_value = partial_value + value
+    if abs(new_value) <= 1e-6
```
rrsadykov, line 91: Why tolerance is hard-coded? We cannot use `Coluna.TOL` here?

src/MathProg/variable.jl
```diff
@@ -8,7 +8,7 @@ mutable struct VarData <: AbstractVcData
     inc_val::Float64
     is_active::Bool
     is_explicit::Bool
-    is_fixed::Bool
+    is_in_partial_sol::Bool
```
rrsadykov, line 11: I am not sure we need this field

src/MathProg/MOIinterface.jl
guimarqu: Because cost may change.

src/MathProg/variable.jl
guimarqu: Yes we need it to show warnings when setting current bounds. This is temporary at the moment.

PR #1069, opened 2023-09-21
First version of presolve 

guimarqu:


rrsadykov:
In the presolve algorithm, we first copy the current partial solution to `PresolveFormRepr` in `create_presolve_reform()`, and then we add this partial solution again to the formulation in `update_reform_from_presolve!()`. Thus, the partial solution is "doubled" each time the presolve algorithm is called. I have corrected this issue in a crude way (I do not copy current partial solution to `PresolveFormRepr`), we need to check whether this is the right way to do it.

rrsadykov (approved):
Ok for me as the first version of presolve, fixes issue #1061.

PR #1076, opened 2023-09-25
Run conquer algorithm in TreeSearch only if `conquer_output` of the node is empty

rrsadykov:
This pull request is a way to ensure that the conquer is not run again needlessly 1) after strong branching 2) at the root of the diving. After running the conquer algorithm, its output (`OptimizationState`) is put to the node so that the divide algorithm can use it even if it is not run immediately after. 

In the future, we would need to define two conquer algorithms for the branch-and-bound (one for the root and one for other nodes), as for the moment we always rely on strong branching to do complete evaluation of the candidate branches. Sometimes, we may want to do a "partial" evaluation in the strong branching. Then, we would need to define an empty conquer algorithm which does nothing and returns nothing. In the when it is used, we would rely on the node to already  possess a conquer output, which we then pass to the divide algorithm.

### DIFF
--- a/src/Algorithm/branching/branchingalgo.jl
+++ b/src/Algorithm/branching/branchingalgo.jl
@@ -274,7 +274,7 @@ end
 
 function Branching.eval_child_of_candidate!(child, phase::Branching.AbstractStrongBrPhaseContext, ip_primal_sols_found, env, reform, input)    
     child_state = OptimizationState(getmaster(reform))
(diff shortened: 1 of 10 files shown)


rrsadykov:
Now, `Node` and `SbNode` are the same. We should remove `SbNode`?

guimarqu (approved):

PR #1137, opened 2025-07-16
Fix label in mermaid diagram

blegat:
Mermaid interprets `o-->` as an arrow starting with a dot, see https://mermaid.js.org/syntax/flowchart.html#multi-directional-arrows. Weirdly enough, it actually does not actually show this dot but it consumes the `o` and we are left with only `n` in the label. Adding this space makes sure that the label is `no`.


rrsadykov:
Dear @blegat,
Thank you for the contribution!
We have just updated the CI: https://github.com/atoptima/Coluna.jl/pull/1140.
Can you please merge the updated master branch to your PRs, so that we can run the CI for them.

PR #1138, opened 2025-07-16
Fix pricing tutorial when no job is assigned

blegat:
Fixes the following errors I got when running the example:
```
ERROR: LoadError: MethodError: no method matching submit(::Model, ::BlockDecomposition.PricingSolution{…}, ::Float64, ::Vector{Any}, ::Vector{…})
```
and
```
ERROR: LoadError: ArgumentError: reducing over an empty collection is not allowed; consider supplying `init` to the reducer
```


blegat:
After https://github.com/atoptima/Coluna.jl/pull/1139 and this PR, the example is working (test with both with GLPK and HiGHS)

